PR #14, opened 2018-07-19
Update service_IAM_role.md

geremyCohen:
Cleaned the document a bit by removing explicit reference to the policies and instead linking to them.  Point to a single source of creating the role via the console, and include an alternative method of creating the role via CloudFormation.  CloudFormation script provided by @palsarma


nrdlngr:
Thanks for your contribution. I like that your method uses the AWS Management Console to show the policies and that you added an AWS CloudFormation example. Would you mind doing a clean rebase and resubmitting so that there are no conflicts?

PR #15, opened 2018-07-20
tensechange.md

otoolema:
The current document wording: NodeGroupName: Enter a name for your node group that is included in your Auto Scaling node group name. makes it sound like these resources should already exist. Changing the tense to reflect that this parameter will apply to resources which will be created by the template.

*Issue #, if available:*

*Description of changes:*


nrdlngr:
Thanks for your contribution.

PR #23, opened 2018-10-25
Add instructions on how to open token window

glenngillen:
The token input modal doesn't appear to open by itself in the current version on first visit to the dashboard, we had to manually go hunting around on how to find it.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/doc_source/dashboard-tutorial.md
+++ b/doc_source/dashboard-tutorial.md
@@ -170,6 +170,8 @@ Now that the Kubernetes dashboard is deployed to your cluster, and you have an a
 
 1. Open the following link with a web browser to access the dashboard endpoint: [http://localhost:8001/api/v1/namespaces/kube\-system/services/https:kubernetes\-dashboard:/proxy/](http://localhost:8001/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/)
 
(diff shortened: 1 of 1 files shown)


nrdlngr:
Interesting. Every time I've created a new cluster and followed this procedure, I've been redirected to the token input model. I wonder if one of us has a browser setting that is giving us a different experience.

What browser are you using? I'd like to try to replicate your issue, and then we'll probably have to come up with some text that works for both scenarios.

nrdlngr:
On second thought, if we change the link to http://localhost:8001/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/#!/login, then it should force the modal.

Can you try that, and if it works, you can change your pull request to use that link and I'll merge it so that you still get contribution credit.

Thanks!

nrdlngr:
I've confirmed that changing the dashboard link to http://localhost:8001/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/#!/login forces the login modal. This change has been published now.

glenngillen:
Belated thank you for following up on this. I'd torn down my environment and didn't get a chance to replicate. Much appreciation on the follow through. A++++, would issue again 🎉🎉🎉😉

PR #27, opened 2018-12-20
Official URL change

mrichman:
The k8s dashboard YAML URL changed 7 hours ago.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/doc_source/dashboard-tutorial.md
+++ b/doc_source/dashboard-tutorial.md
@@ -20,7 +20,7 @@ Use the following steps to deploy the Kubernetes dashboard, `heapster`, and the 
 1. Deploy the Kubernetes dashboard to your cluster:
 
    ```
-   kubectl apply -f https://raw.githubusercontent.com/kubernetes/dashboard/master/src/deploy/recommended/kubernetes-dashboard.yaml
+   kubectl apply -f https://raw.githubusercontent.com/kubernetes/dashboard/master/aio/deploy/recommended/kubernetes-dashboard.yaml
    ```
 
    Output:


mrichman:
More info from the k8s team:

> As for the yaml relocation. Any kind of documentation that points directly to the master branch of any project is destined to fail at some point. We will be updating documentation to always point to the latest release, but we will not use master as a reference anymore. Master branch can be unstable and can contain breaking changes. We did not just migrate the angular, we have  rewritten pretty much the whole application including build pipeline, etc. That is why there was complete project refactoring and some paths has had to change.

nrdlngr:
Thanks for your pull request. We've opted to use the latest release branch instead of master. I've updated the dashboard tutorial accordingly and tested that it works for both 1.10 and 1.11 Amazon EKS clusters.

RothAndrew:
@mrichman source?

> More info from the k8s team:
> 
> > As for the yaml relocation. Any kind of documentation that points directly to the master branch of any project is destined to fail at some point. We will be updating documentation to always point to the latest release, but we will not use master as a reference anymore. Master branch can be unstable and can contain breaking changes. We did not just migrate the angular, we have  rewritten pretty much the whole application including build pipeline, etc. That is why there was complete project refactoring and some paths has had to change.

mrichman:
@RothAndrew https://kubernetes.slack.com/archives/C09R1JVU6/p1545386237005600

mrichman:
@nrdlngr Yes, this is definitely the best approach. You might also consider updating the screenshot in the docs. The UI is slightly different now.

PR #31, opened 2019-01-11
Changes in the private subnet consideration

badaldavda8:
This is because for internal ELB auto subnet discovery, both tags are used -
kubernetes.io/role/internal-elb	1
kubernetes.io/cluster/<cluster-name>	shared

Since as per code, first kubernetes.io/cluster/<cluster-name> is checked and then kubernetes.io/role/internal-elb is checked.

If kubernetes.io/cluster/<cluster-name> is not mentioned, then internal ELB is created on Public Subnets.
https://github.com/kubernetes/kubernetes/issues/29298#issuecomment-356826381

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/doc_source/network_reqs.md
+++ b/doc_source/network_reqs.md
@@ -45,3 +45,6 @@ Private subnets in your VPC should be tagged accordingly so that Kubernetes know
 | Key | Value | 
 | --- | --- | 
 |  `kubernetes.io/role/internal-elb`  |  `1`  | 
+|  `kubernetes.io/cluster/<cluster-name>`  |  `shared`  | 
++ **Key**: The *<cluster\-name>* value matches your Amazon EKS cluster's name\. 
++ **Value**: The `shared` value allows more than one cluster to use this VPC\.


M00nF1sh:
Hi, 
I suggest moving the  "Private Subnet Tagging Requirement for Internal Load Balancers" as a level 3 header under "Subnet Tagging Requirement" instead.  Since the "kubernetes.io/cluster/" is already expressed in "Subnet Tagging Requirement". 
Also, we can add "Public Subnet Tagging" as a sub-item under "Subnet Tagging Requirement" as well.

nrdlngr:
@M00nF1sh, I agree with your suggestion. I think the public/private considerations should fall under the subnet requirements, since the cluster tag is required for ALL subnets.

nrdlngr:
@badaldavda8, thanks for your interest in improving our documentation. I've opted instead to move the load balancer requirements under the subnet tag section, and I've made it more obvious that the cluster tag should apply to all subnets.

PR #32, opened 2019-01-18
Added example for deploying private EKS nodes with the ability to deploy Public facing ELB/ALB/NLB resources.

danmanners:
*Description of changes:*

The documentation all but flat out stated that you cannot deploy private nodes and then deploy public resources. This is not true given the appropriate setup. The code below shows how to do just this.

### DIFF
--- a/doc_source/network_reqs.md
+++ b/doc_source/network_reqs.md
@@ -8,7 +8,7 @@ Internet\-facing load balancers require a public subnet in your cluster\.
 The subnets that you pass when you create the cluster influence where Amazon EKS places elastic network interfaces that are used for the control plane to worker node communication\.
 
 It is possible to specify only public or private subnets when you create your cluster, but there are some limitations associated with these configurations:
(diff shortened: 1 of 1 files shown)


nrdlngr:
Hi @danmanners, thanks for your submission, and I'm really sorry for the delayed response.

When you create your Amazon EKS cluster, you should specify all of the subnets that you intend to use in your VPC, not just the subnets that you want to use for worker nodes. When you do that, Amazon EKS automatically tags your subnets properly.

So I'm hesitant to add your procedure for manually creating route tables (that should just be part of your VPC creation workflow). And you wouldn't need to manually tag the subnets, because that would be done for you if you specify them when you create the cluster.

If you want your worker nodes to only reside in the private subnets, only specify the private subnets when you launch your worker nodes.

However, I see your point that the docs can be misleading about what you can and can't do with subnets that aren't specified when you create the cluster. So I am open to merging your first code block (line 11) if you want to update your PR to remove the second code block.

nrdlngr:
Closing due to inactivity. Feel free to reopen this pull request or create a new one if you address my comments. Thanks!

PR #33, opened 2019-01-22
Correcting instructions to allow pods use secondary VPC CIDR

anshrma:
Running what is mentioned in docs results in error as follows, 

kubectl apply -f ENIConfig.yaml
error: error validating "ENIConfig.yaml": error validating data: ValidationError(CustomResourceDefinition.spec.names): unknown field "scope" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionNames; if you choose to ignore these errors, turn validation off with --validate=false

Following works

kubectl apply -f ENIConfig.yaml --validate=false
customresourcedefinition.apiextensions.k8s.io "eniconfigs.crd.k8s.amazonaws.com" configured

### DIFF
--- a/doc_source/cni-custom-network.md
+++ b/doc_source/cni-custom-network.md
@@ -55,7 +55,7 @@ This feature requires [Amazon VPC CNI plugin for Kubernetes](https://github.com/
    1. Apply the file to your cluster with the following command:
 
       ```
-      kubectl apply -f ENIConfig.yaml
+      kubectl apply -f ENIConfig.yaml --validate=false
       ```
 
 1. Create an `ENIConfig` custom resource definition for each subnet in which your pods reside\.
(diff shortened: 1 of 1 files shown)


taylorb-syd:
Instead of turning of validation, which is a terrible idea, why don't we instead fix the problem? Which is the errant `scope` under `names`. Definition should be:

```
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: eniconfigs.crd.k8s.amazonaws.com
spec:
  scope: Cluster
  group: crd.k8s.amazonaws.com
  version: v1alpha1
  names:
    plural: eniconfigs
    singular: eniconfig
    kind: ENIConfig
```

nrdlngr:
As @taylorb-syd  pointed out, the validation error was due to a typo in the ENI config example. I've corrected that error, so I'm closing this PR.

@anshrma, thanks for your interest in improving our documentation!

PR #34, opened 2019-02-01
Improved Documentation Regarding CNI Custom Network re 1.4.0 release

taylorb-syd:
*Issue #, if available:* N/A

*Description of changes:* With the soon to be released changed for [1.4.0 of the CNI Plugin](https://github.com/aws/containers-roadmap/issues/149) the behavior of CNI Custom Networks [has changed](https://github.com/aws/amazon-vpc-cni-k8s/pull/280).

This pull request expands on how customers might use this, and a common use case involving the _failure-domain.beta.kubernetes.io/zone_ label.

Since the PR is dependent on release 1.4.0 of the CNI Plugin, I strongly recommend against merging this until 1.4.0 is released and contingent on the linked PR being included.


nrdlngr:
Thanks so much for your contribution here, @taylorb-syd. The original content has changed since your submission, so I had to do some refactoring. The service team suggested that we stick to just labels instead of offering steps for both labels and annotations (to avoid confusion). I pulled in your procedure for automatically applying an ENIConfig per the node's Availability Zone, however, because that content did not already exist in our docs.

PR #54, opened 2019-09-05
Example snippet for downloading metrics server

elphastori:
Added an example snippet for downloading a metrics server release using curl on Linux.

This is the natural next step when following this tutorial while connected to a headless server but, there are several pitfalls as shown in this article https://echorand.me/posts/curl-download-redirect/


nrdlngr:
This is a fair point: "the jq utility, although quite common, does not come preinstalled on some major Linux distributions and certain mac versions so I had to spend some time getting the permission to install to it and then finally installing it"

I think this is a reasonable addition that will help in some use cases like you describe. Could you modify the curl command so that it works with the latest release? I need to update this topic to point to the latest release anyways, and I can make those changes retroactively.

elphastori:
I have modified the curl command and updated the release in the other steps. Thank you for your help!

PR #56, opened 2019-09-15
Fix log stream name of EKS control plane log

shibataka000:
*Issue #, if available:*

No issue.

*Description of changes:*

Fix log stream name of EKS control plane log. In my environment, controller manager's (`controllerManager`) log stream starts with `kube-controller-manager-`, and scheduler's (`scheduler`) log stream starts with `kube-scheduler-`.


nrdlngr:
Good catch! Thanks for your contribution.

PR #57, opened 2019-09-16
Fix instructions for checking coredns version

lachlancooper:
*Description of changes:* The result of `cut` only has two fields, so attempting to show the third results in no output.

### DIFF
--- a/doc_source/update-cluster.md
+++ b/doc_source/update-cluster.md
@@ -92,7 +92,7 @@ This procedure only works for clusters that were created with `eksctl`\.
 1. Check the current version of your cluster's `coredns` deployment\.
 
    ```
-   kubectl describe deployment coredns --namespace kube-system | grep Image | cut -d "/" -f 3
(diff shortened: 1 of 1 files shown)


nrdlngr:
Thanks for your suggestion.

I ran both the command in the current docs and your suggested command, and here is the resulting output:
```
ericn@ericn-mac:~$ kubectl describe deployment coredns --namespace kube-system | grep Image | cut -d "/" -f 3
coredns:v1.3.1
ericn@ericn-mac:~$ kubectl describe deployment coredns --namespace kube-system | grep Image | cut -d "/" -f 2
eks
```

The current documented command returns the expected output, `coredns:v1.3.1`. Are you running this command against an Amazon EKS cluster?

lachlancooper:
I was running this against an EKS cluster that was upgraded from version 1.10, with CoreDNS installed manually. Apparently we are using the regular `coredns/coredns` image rather than the one published by AWS, which is why the command didn't produce the expected output. My apologies.

PR #59, opened 2019-10-03
PodPreset is not yet available on EKS

starchx:
To avoid confusion, I suggest to either adding a note (as this PR), or remove PodPreset from the context.

*Issue #, if available:*

*Description of changes:*
The doc was suggesting to use PodPreset as a workaround to configure the Pod as an alternative to the Pod Identity Webhook, however, PodPreset is not yet in Beta, and not enabled on EKS yet: https://github.com/aws/containers-roadmap/issues/151


nrdlngr:
Thanks for your contribution! I'll merge this in for now so you get credit for your submission, but I'll probably remove the mention of PodPresets until they are supported on Amazon EKS to avoid confusion.

PR #63, opened 2019-10-24
Add Tagging Requirement for Worker Node Security Group

starchx:
As per reported issue on here: https://github.com/kubernetes/kubernetes/issues/84287
Worker node security group requires a cluster tag so Cloud Controller Manager knows which security group to add or remove the ingress rules. Otherwise, it will cause load balancer security group failed to get removed.

*Issue #, if available:*
84287

*Description of changes:*
Add worker node security group tagging requirement.


doc_source/network_reqs.md
```diff
@@ -71,4 +71,14 @@ Public subnets in your VPC may be tagged accordingly so that Kubernetes knows to
 
 | Key | Value | 
 | --- | --- | 
-|  `kubernetes.io/role/elb`  |  `1`  | 
\ No newline at end of file
+|  `kubernetes.io/role/elb`  |  `1`  | 
+
+## Worker Node Security Group Tagging Requirement<a name="worker-node-security-group-tagging"></a>
+
+When you create LoadBalancer type Kubernetes service, one of the security groups for the worker node should be tagged so that Kubernetes Cloud Controller Manager knows which security group to add or remove ingress rule for load balancer security group:
```
nckturner: Can we specify that this is optional if there is only one security group on the worker node?

/cc @M00nF1sh

M00nF1sh: yes, we should be clear on two fact:
1. The tag on securityGroup is optional if you only have one securityGroup on your instance(primary ENI)
2. The tag is required on **exactly one** SecurityGroup if you have multiple securityGroups on your instance(primary ENI)

starchx: Sorry for delay. Thanks for the feedback - 100% agreed. Since this tagging requirement is related to load balancer, I believe it is more suitable to put on load-balancing.md page. 
Please review this PR instead: https://github.com/awsdocs/amazon-eks-user-guide/pull/79
Closing this PR.

PR #69, opened 2019-11-06
Added one more reason to Worker Nodes Fail to Join Cluster

sudheerchamarthi:
If the Node is not owned by cluster, Node will fail to join the cluster. Adding this step here will help newbies.

*Issue #, if available:*

*Description of changes:*


nrdlngr:
Thanks for your contribution! I'll apply a few edits after merging this, but you are correct, this is a requirement.

PR #71, opened 2019-11-24
Add STS regional endpoint

starchx:
IAM Role for Service Account supports STS regional endpoint, but it is not documented. This is to add documentation of how to configure it to work with STS regional endpoint.

*Description of changes:*

As shown here (https://github.com/aws/amazon-eks-pod-identity-webhook/blob/2fd4becd434f92edab01cecee30ea4b12af93ab7/hack/example-app.yaml), and tested by myself, IAM Role for Service Account can work with regional STS endpoint. This suits some customers who don't want to use global STS endpoint. 

### DIFF
--- a/doc_source/iam-roles-for-service-accounts-technical-overview.md
+++ b/doc_source/iam-roles-for-service-accounts-technical-overview.md
@@ -182,3 +182,27 @@ web_identity_token_file = /var/run/secrets/eks.amazonaws.com/serviceaccount/toke
 role_arn=arn:aws:iam::111111111111:role/account-a-role
 ```
 To specify chained profiles for other AWS SDKs, consult their documentation\.
+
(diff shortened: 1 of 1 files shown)


micahhausler:
I would not recommend changing the documented string to a regional STS domain name. The audience string value is arbitrary, really just meant to indicate who is the consumer of the token. The value itself `sts.amazonaws.com` is not ever called. 

I tried setting the `AWS_REGION` env var with the aws cli, but it didn't seem to have any effect on switching from the global endpoint. You'd have to open an issue with the SDK teams (most likely on https://github.com/aws/aws-cli) to find out how to configure this

starchx:
You are right. The only thing it matters is that the token's audience needs to be on the audience list from the IAM OIDC provider. Thanks. Closing PR.

PR #84, opened 2020-01-07
Adding note into Fargate Considerations for selected subnets

0xlen:
*Issue #, if available:*

When using EKS on Fargate, the private subnets should have capability to connect with internet otherwise Pod will stuck at pending state:

```
$ kubectl get pods --watch
NAME                        READY   STATUS    RESTARTS   AGE
demo-app-6dbfc49497-qjs7r   0/1     Pending   0          15m
```

**Reference**
- https://github.com/aws/containers-roadmap/issues/666

*Description of changes:*


jimdial-aws:
Thank you. I'm happy to change this text, but I'd prefer to align it with this text "Pods running on Fargate are only supported on private subnets (with NAT gateway access to AWS services, but not a direct route to an Internet Gateway), so your cluster's VPC must have private subnets available." from the [Fargate Profile](https://github.com/awsdocs/amazon-eks-user-guide/blob/master/doc_source/fargate-profile.md) doc. Would you mind changing your text to that and I'll merge the PR?

PR #97, opened 2020-03-04
Fix internal loadbalancer annotation

aliusmiles:
*Description of changes:*

Fix internal load balancer annotation based on  [k8s docs](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer):
```
[...]
metadata:
    name: my-service
    annotations:
        service.beta.kubernetes.io/aws-load-balancer-internal: "true"
[...]
```


M00nF1sh:
/lgtm
Thanks for fixing this

PR #100, opened 2020-03-10
Update Fargate documentation

dapseen:
*Description of changes:*

Line 49 returns null, this is due to typo in the query


jimdial-aws:
If that command returns null for you, it could be due to you running an older version of the EKS CLI. If you [install the latest version](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-install.html), you should find that a value is returned.

dapseen:
Thanks for that. I think this should be included in the documentation

PR #107, opened 2020-03-28
Update create-service-account-iam-policy-and-role.md

dongnorm:
Update trust relationship json file to be used with AWS CLI. The original version that uses OIDC sub does not work any more per my testing.

*Issue #, if available:*

*Description of changes:*


jimdial-aws:
I removed the steps that set the variables, but left line 180 as-is.

PR #134, opened 2020-05-25
Clarify documentation of what sunets will be used for internal load balancers

dianibar-work:
*Issue #, if available:*

I found that if I create a internal load balancer for an EKS cluster, where not private subnets are tagged with “key: kubernetes.io/role/internal-elb Value: 1”, but there is at least one subnet associated with the cluster in a specific an Availability zone, by using the tag “key: kubernetes.io/cluster/qa-eps-eks, value: shared”. EKS it will create the load balancer choosing one of the associated private subnets in lexicographically order. Moreover if the Availability Zone has not a private subnets associated with the cluster,  a public subnet associated to the cluster will be added to the Load balancer instead.

This behavior is described in the algorithm that chooses the Load Balancers subnets [1]. Also I replicate this behavior in my Kubernetes cluster.

References:

[1]

*Description of changes:*

I would like to clarify how the subnets are chosen when internal Load Balancer is created. I will choose a private subnet per availability zone where there is a subnet associated to the cluster although no subnets has a tag “key: kubernetes.io/role/internal-elb Value: 1”

### DIFF
--- a/doc_source/load-balancing.md
+++ b/doc_source/load-balancing.md
@@ -27,7 +27,9 @@ You must tag the public subnets in your VPC so that Kubernetes knows to use only
 | --- | --- | 
 | `kubernetes.io/role/elb` | `1` | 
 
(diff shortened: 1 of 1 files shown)


jimdial-aws:
Closing this due to a merge conflict and a significant update coming to this document soon that should address this and other issues. Thank you for your submission.

PR #148, opened 2020-06-08
Add target type requirement for Fargate

lnregalias:
*Issue #, if available:*

*Description of changes:*

Currently (from testing), the target type must be set to IP mode when using Fargate pods for correct operation - otherwise the ALB will return HTTP 503 and have no registered targets in the instance mode target group.

The 2048 example for Fargate further below on the same page also instructs users to do this by adding the `alb.ingress.kubernetes.io/target-type: ip` annotation - this requirement should be highlighted for better visibility.


jimdial-aws:
This has a merge conflict but I have added your text. Thank you for the submission.

PR #150, opened 2020-06-10
Update CNI IAM role configuration walkthrough

couralex6:
The user still needs to create an OIDC identity provider and a Kubernetes service account even if they do not need to upgrade the CNI version. Additionally, the user needs to restart the CNI plugin for the changes to come into effects.

*Description of changes:*

* Moved steps 2.a and 2.b to outside of step 2. 
* Added instructions to restart the CNI plugin pods.


jimdial-aws:
This document was deprecated. The new version is [here](https://github.com/awsdocs/amazon-eks-user-guide/blob/master/doc_source/cni-iam-role.md). I think it addresses your concerns, but feel free to create a new PR against the new document if you disagree. Thank you.

PR #152, opened 2020-06-13
Add next step links for network policy

lxpollitt:
Add next step links to getting started with EKS guides for getting started with network policy.


jimdial-aws:
Thanks for the submission. I'm not going to add this link to the next steps because while the current next steps are basic tasks that a new user would likely want to do, I think that applying network policy might be more of an advanced task that I would expect most users not to do right away.

lxpollitt:
Securing your cluster should not be an after thought,  though I appreciate it is for many.  I suggested this PR after several people told me that they did not realize that EKS supports network policy.  And I didn't think there was a significant impact on people who aren't interested in this, they can skip it, just like many people will skip helm or dashboard etc.

PR #160, opened 2020-06-22
Add IAM permissions when creating service-linked roles at first.

hiraken-w:
The documentation for service-linked roles does not specify the IAM permissions required for creation.
So I added an IAM permissions based on my verification.
Could you please confirm this?


jimdial-aws:
You're right, the text does not include the permissions required to create a service linked role. It does however, reference, and link to the IAM document that does, and it is the authoritative source for what permissions are required, which is why I won't add the addition to the document. Thank you for the submission though.

Issue #166, opened 2020-06-30
Minimum client TLS version is incorrect.


raonitimo:
Document says Clients must support Transport Layer Security (TLS) 1.0 or later. But tests with curl, https://github.com/rbsec/sslscan, and https://www.ssllabs.com/ssltest/ on an EKS cluster endpoint show the minimum required TLS version is 1.2.

PR #167, opened 2020-06-30
Fix infrastructure-security.md

raonitimo:
Fixes issue #166 

Kubernetes default is VersionTLS12: https://github.com/kubernetes/kubernetes/blob/release-1.16/staging/src/k8s.io/component-base/cli/flag/ciphersuites_flag.go#L101

Unless we set --tls-min-version on kube-apiserver (https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/).

Description of changes:
Remove mention to 1.0 or later, and define clients must support Transport Layer Security TLS 1.2 or later.


jimdial-aws:
Amazon EKS does not require 1.2, though it is recommended, as stated in the current document. Thank you for using Amazon EKS!

PR #170, opened 2020-07-09
Make a clear comment to not use `force_tcp` for CoreDNS

mogren:
*Issue #, if available:* https://github.com/aws/amazon-vpc-cni-k8s/issues/595

*Description of changes:*

We have seen multiple issues around DNS because of an underlying limitation in EC2 when it comes to DNS lookups over TCP. The suggested work-around is to not use `force_tcp`. See https://github.com/aws/amazon-vpc-cni-k8s/issues/595#issuecomment-649872775 for details.


nithu0115:
Nice! Thanks updating documentation, @mogren

PR #175, opened 2020-07-15
Corrected kube-proxy image update section

shyamjos:
Userguide is suggesting user to change the kube proxy image repository URL from `602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/kube-proxy:v1.16.12` to user's region and account ID but changing repository URL to user's aws account id will fail with an error since user does not have an  eks/kube-proxy image in their repo by default.

`xxxxxxxxx.dkr.ecr.eu-west-1.amazonaws.com/eks/kube-proxy:v1.16.12": rpc error: code = Unknown desc = Error response from daemon: repository xxxxxxxxx.dkr.ecr.eu-west-1.amazonaws.com/eks/kube-proxy not found: name unknown: The repository with name 'eks/kube-proxy' does not exist in the registry with id 'xxxxxxxxx'`

But if the user only change the region field and keep the account id same as in the userguide it will work, so I think it's a typo/mistake made by the author

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/doc_source/update-cluster.md
+++ b/doc_source/update-cluster.md
@@ -192,7 +192,7 @@ The cluster update should finish in a few minutes\.
        kube-proxy=602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/kube-proxy:v1.17.7
    ```
 
-   Your account ID and region may differ from the example above\.
+   Your region may differ from the example above\.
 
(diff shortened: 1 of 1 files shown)


lavoiml:
Hi shyamjos. Thank you for your contribution! What the author is saying is that the account ID the customer is using may be different from the provided example. With that in mind, I'm going to go ahead and close this PR. Thank you for being an active member of our community!

PR #178, opened 2020-07-20
Added that encryption is possible for Managed Node Groups

jasonumiker:
I find that many do not know that you can encrypt Managed Mode Groups via EBS default encryption - wanted to make that clearer in the documentation.


jimdial-aws:
Closing this since you can now deploy encrypted EBS volumes for nodes if you deploy the nodes with a launch template.

PR #193, opened 2020-08-17
Adds instructions on how to fix AccessDenied errors with managed node…

mmerkes:
… groups

*Issue #, if available:*

N/A

*Description of changes:*

Getting `AccessDenied` errors during managed node groups operations is one of the more common error types, and it can be fixed with a few `kubectl` commands. I've added a section on how to fix the issue in the most common cases. I'm open to feedback on how to improve the format of the section.


doc_source/troubleshooting.md
```diff
@@ -76,9 +76,141 @@ If your managed node group encounters a health issue, Amazon EKS returns an erro
...
 
+### Fixing AccessDenied errors for managed node groups
+
+The most common cause of `AccessDenied` errors when performing operations on managed node groups is missing the `eks:node-manager` ClusterRole and/or ClusterRoleBinding. Amazon EKS sets up these resources in your cluster as part of onboarding with managed node groups, and these are required for managing the node groups.
+
+You can verify the cause and resolve the issue with a few `kubectl` commands.
+
+**Expected ClusterRole and ClusterRoleBinding**
+
+The ClusterRole and ClusterRoleBinding are subject to change over time, but they should look similar to the following:
```
vishalkg, line 90: Instead, should we put it as, what "should" be there? Saying, "are subject to change over time" might be a little confusing, since the permissions to these roles pretty scooped down

mmerkes: We can. My only concern is that at some point in time, we could theoretically lock them down further or need to add additional permissions. I don't know how realistic that is and maybe it will be less confusing to not go into that, but I think it's worth discussing what makes sense.

PR #200, opened 2020-09-10
Update CNI URLs to use git tag instead of branch

mogren:
*Issue #, if available:*
https://github.com/aws/amazon-vpc-cni-k8s/issues/1206

*Description of changes:*

Update the docs to point to CNI v1.7.2. This should be the default CNI for EKS later in September, so do not merge yet.

Note that now that we use one branch per minor release (release-1.7, release-1.8 and so on) we rely on git tags for the patch releases. This PR updates the sample URLs to use the git tags instead of branch names..


jimdial-aws:
Thanks for your submission.

PR #208, opened 2020-09-28
patch cluster-autoscaler deployment to run on bottlerocket worker nodes

muhyousri:
*Issue #, if available:*

The ca-bundle.crt file is not in this path /etc/ssl/certs/  in bottlerocket AMI , so deploying this will cause the error 

mounting "/etc/ssl/certs/ca-bundle.crt" to rootfs at "/run/containerd/io.containerd.runtime.v2.task/k8s.io/cluster-autoscaler/rootfs/etc/ssl/certs/ca-certificates.crt" caused: not a directory: unknown

*Description of changes:*

Added an optional patch command for using with bottlerocket nodes , the file is located in this path :
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/factory/etc/pki/tls/certs/ca-bundle.crt


jimdial-aws:
Thanks for the submission. Sorry for the delayed reply. Closing due to a merge conflict, but I did incorporate your `kube-proxy` changes into the update cluster document. I didn't incorporate the Bottlerocket changes because I believe that [this pull request](https://github.com/bottlerocket-os/bottlerocket/pull/1207) may make them unnecessary. Feel free to re-open if that PR does not address your concern.

PR #214, opened 2020-10-16
Suggest the simpler "annotate" command instead of a JSON patch

jpetazzo:
This is super minor, so feel free to disregard if it doesn't fit with the
general AWS style guide; but `kubectl annotate` is simpler than the
longer `kubectl patch`.

I also replaced `<gp2>` with `gp2` for consistency. (I understand that
`<>` is used in some places in the docs as a placeholder, but since
all other places in this doc use `gp2`, I thought it would avoid issues
if someone copy-pastes the command blindly.)

*Issue #, if available:*

*Description of changes:*


jimdial-aws:
Thanks for the submission!

PR #223, opened 2020-10-20
Make the CoreDNS ConfigMap check more precise

dbirks:
*Issue #, if available:*

*Description of changes:*

The current command will give you a false positive, even after removing
the `upstream` line, because it's still in the yaml under `last-applied-configuration`.

Example:
  Even after removing the `upstream` config option, I expect to get an
  empty response, but I see this instead:
```
kubectl get configmap coredns -n kube-system -o yaml | grep upstream
      {"apiVersion":"v1","data":{"Corefile":".:53 {\n    errors\n    health\n    kubernetes cluster.local in-addr.arpa ip6.arpa {\n      pods insecure\n      upstream\n      fallthrough in-addr.arpa ip6.arpa\n    }\n    prometheus :9153\n    forward . /etc/resolv.conf\n    cache 30\n    loop\n    reload\n    loadbalance\n}\n"},"kind":"ConfigMap","metadata":{"annotations":{},"labels":{"eks.amazonaws.com/component":"coredns","k8s-app":"kube-dns"},"name":"coredns","namespace":"kube-system"}}
```

This PR adds `-o jsonpath='{$.data.Corefile}'` to narrow it down to just
the actual coredns config before passing it off to grep. No more false
positives! :)

### DIFF
--- a/doc_source/update-cluster.md
+++ b/doc_source/update-cluster.md
@@ -64,7 +64,7 @@ Update the cluster and Kubnernetes add\-ons\.
    1. Check to see if your CoreDNS manifest has the line\.
 
       ```
-      kubectl get configmap coredns -n kube-system -o yaml |grep upstream
+      kubectl get configmap coredns -n kube-system -o jsonpath='{$.data.Corefile}' | grep upstream
       ```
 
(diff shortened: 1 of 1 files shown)


dbirks:
Oh and can you add the label `hacktoberfest-accepted` as well, if it looks good to you? Thank you!

jimdial-aws:
Thank you for the submission!

PR #228, opened 2020-10-28
Update create-cluster.md

hassanSRE:
*Issue #, if available:*

*Description of changes:*

(to avoid such ambiguity) instead of "Select", "Unselect" would be better phrase since the subnet at this step is already preselected and we should un-tick/unselect the one that we do not want,


lavoiml:
Hi hassanSRE! Thank you for your contribution! I've incorporated your suggested changes into the docs. I'm closing this PR instead of merging it, because I'm doing one big update incorporating multiple PRs. Check the docs to see your contribution reflected there. Thank you for being an active member of our community!

PR #234, opened 2020-10-30
Mention additional IAM policies required for upgrade

kishorj:
*Issue #, if available:*
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1577
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1578

*Description of changes:*
When upgrading from AWS ALB ingress controller to the new AWS Load Balancer controller, additional IAM policies need to be granted for controller to work properly after migration. Mention the step in the AWS documentation.


jimdial-aws:
Thanks for the submission!

PR #240, opened 2020-11-08
Update fargate node affinity Rule 

muhyousri:
This is causing alot of confusion for people that created old clusters (>=1.13) and been following the regular upgrade path . when they start using fargate , the kube-proxy daemonset will try to launch new pods on all new fargate nodes because the node affinity is only added by default to kube-proxy starting nodes created with v1.14 , alot of pending pods is usually cause for concerns where it can be avoided

The same issue exist in aws-node Daemonset but it's usually upgraded  by applying the new manifest completely not just by changing the image version like kube-proxy and the affinity is added there along time ago

### DIFF
--- a/doc_source/update-cluster.md
+++ b/doc_source/update-cluster.md
@@ -213,6 +213,22 @@ The cluster update should finish in a few minutes\.
                             - amd64
                             - arm64
       ```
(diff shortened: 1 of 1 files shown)


jimdial-aws:
Closing this one since it duplicates information in one of your other [PR](https://github.com/awsdocs/amazon-eks-user-guide/pull/208/)s.

PR #246, opened 2020-11-15
Set the backup extension for sed's in-place argument

ankon:
The `-i` option requires an extension argument on macOS

As there doesn't seem to be a reasonably portable variant that doesn't produce backups (see https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux) this simply lets sed produce a .bak file.


jimdial-aws:
I updated this to sed -i.bak, which I believe also works on macOS. Let me know if not. Thanks.

PR #253, opened 2020-12-01
Update kube-proxy version tag

tde908:
Updated the `kube-proxy` version tag from `1.15.12` to `1.15.12-eksbuild.1` as the tag naming convention has changed.


jimdial-aws:
Thanks for your submission. That table lists versions, not tags, so I won't merge this change. A step later in the document explains how to pull the image, and it does have the tag you added in it. Thanks for using Amazon EKS!

PR #254, opened 2020-12-02
Fix meaning of CLB targets type limitation to opposite

mtilson:
*Issue #, if available:*
N/A

*Description of changes:*
According to the previous limitation:
> You can only use IP targets with NLB. You can't use IP targets with CLBs.


jimdial-aws:
This change conflicts with the base file now, but the change has been made. Thanks for your submission!

PR #257, opened 2020-12-03
Update View Workload Prerequisites

SypherSu:
*Description of changes:*

Modify the first prerequisites that might misleading customer as currently only mention IAM user can view workload.


jimdial-aws:
Thank you for your submission, and for using Amazon EKS!

PR #267, opened 2020-12-10
Update delete-managed-node-group.md

suket22:
Changes to description to better reflect how deleting a managed nodegroup works.

*Issue #, if available:*

*Description of changes:*
The description earlier claimed that that nodes would be terminated one at a time. Updating this description to show that the size of the  Auto Scaling group is set to 0, which leads to AutoScaling group attempting to terminate multiple instances at once.


jimdial-aws:
Thanks for the submission!

PR #269, opened 2020-12-11
Update fargate-logging.md

chaudhary27:
*Issue #, if available:*
NA

*Description of changes:*
Removed <> around region, this is causing LogGroup to not show up in CloudWatch. Thanks to Pierre Beaumier for identifying this.

### DIFF
--- a/doc_source/fargate-logging.md
+++ b/doc_source/fargate-logging.md
@@ -54,7 +54,7 @@ Apply a `ConfigMap` to your Amazon EKS cluster with a `Fluent Conf` data value t
              [OUTPUT]
                  Name cloudwatch_logs
                  Match   *
-                 region <us-east-1>
+                 region us-east-1
                  log_group_name fluent-bit-cloudwatch
                  log_stream_prefix from-fluent-bit-
                  auto_create_group On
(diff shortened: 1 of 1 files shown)


jimdial-aws:
The brackets are there to denote a replaceable value (and the text a few lines above your proposed change says "Create the namespace. Replace the <example values> (including <>) with your own values.") If you think it's necessary to add a statement like that to every step where it's applicable, we can certainly do that, but I'd prefer not to remove the brackets, because the value is replaceable. Let me know. Thanks.

pmbeaumier:
@jimdial-aws I would then advocate for the value within the bracket to be explicitly wrong `<region-code>` for example.
AWS documentation often uses a different color (red usually) to denotes values that should be changed in configuration samples (https://docs.aws.amazon.com/codebuild/latest/userguide/sample-ecr.html as an example)

jimdial-aws:
You're right that AWS documentation normally does use a different color, but unfortunately, markdown does not. The brackets allow consistency across the two and are commonly used in technical documentation to represent replaceable values. Thanks for your feedback, and for using Amazon EKS!

PR #271, opened 2020-12-16
Add note for SG rules handling in EKS 1.16+

kishorj:
*Issue #, if available:*

*Description of changes:*

Recent EKS platforms for 1.16 and later versions contain the upstream
fix "Use NLB Subnet CIDRs instead of VPC CIDRs in Health Check SG
Rules". Add a note for the changed behavior and the recommendations in
case customers hit the security group rules limit.


jimdial-aws:
Thanks for the submission!

PR #274, opened 2020-12-29
Add aws-observability ConfigMap namesapce requirement and  performanc…

HallBlazzar:
…e considerations to the document.

1. The [blog post](https://aws.amazon.com/blogs/containers/fluent-bit-for-amazon-eks-on-aws-fargate-is-here/) mentions that the `aws-logging` ConfigMap has to be created in a fixed namespace called `aws-observability` to enable Fargate logging. However, the document doesn't mention that. It will cause users mistake that they could create the ConfigMap in arbitrary namespace to enable and customize Fargate logging of the namespace. But in fact, the effect of the ConfigMap is cluster-wised.
2. Add performance consideration to the document to make user know that they should preserve additional resource for Fargate logging while requesting resources for Fargate Pods.

*Issue #, if available:*

*Description of changes:*


jimdial-aws:

I didn't add the text from your PR for "performance considerations", because that information is already in a section with the heading "size considerations", just above the Troubleshooting section.

PR #276, opened 2021-01-04
updated logging troubleshooting with a note on namespace

saiteja313:
*Issue #, if available:* n/a

*Description of changes:*
This pull request will update troubleshooting documentation with a note point on ConfigMap related to Fargate logging configuration for EKS Clusters.


jimdial-aws:

PR #278, opened 2021-01-05
Update EKS fargate logging documenation

sumanthkumarc:
The current EKS fargate log docs mentions steps to create namespace and configmap properly.  But sentences in doc like `Create the namespace. Replace the <example values> (including <>) with your own values.` or steps in pre-requisites tend to make even the seasoned kubernetes user to think that the namespace creation is an example if current namespace doesn't exist. After all, if i'm creating fargate profiles specially for namespaces, one would also tend to think that this logging is namespace specific. The changes i suggested here are clearly mentioned in release blog at  https://aws.amazon.com/blogs/containers/fluent-bit-for-amazon-eks-on-aws-fargate-is-here/ but felt ambiguous in main AWS docs :(


jimdial-aws:

PR #313, opened 2021-02-18
Load balancer tagging is wrong

lachlanbb:
The following is wrong, This will do the opposite of what is intended and is a security issue. 

The following page has the correct configuration. 
https://aws.amazon.com/premiumsupport/knowledge-center/eks-vpc-subnet-discovery/

*Issue #, if available:*

*Description of changes:*


jimdial-aws:
Thank you for catching and bringing it to our attention.

PR #314, opened 2021-02-18
document system:node-proxier clusterrole changes for EKS 1.19

kishorj:
*Issue #, if available:*

*Description of changes:*
kube-proxy included in EKS 1.19 requires read permissions to `endpointslices` resources. If customers upgrade to EKS 1.19 and modify the kube-proxy daemonset, they need to specify additional permissions. 


kishorj:
There is no issue when following the EKS upgrade path. i.e when EKS 1.17 cluster gets upgraded to EKS 1.18 the system:node-proxier role gets modified appropriately.

The issue is apparent only in case the system resources get managed externally. It is up to the end user to ensure that the external component correctly gets the updated resources from the EKS.

PR #318, opened 2021-02-22
Fixes reference to nodegroup CFN template yaml

mmerkes:
*Issue #, if available:*

[AMI issue](https://github.com/awslabs/amazon-eks-ami/issues/621)

*Description of changes:*

The current guidance fails with a `Template format error: YAML not well-formed` because it downloads the HTML page, not the actual yaml. If we link to the GitHub raw link, it works for me.


jimdial-aws:
Thank you!

PR #330, opened 2021-03-04
[Discrepancy on EKS documentation] Creating a Fargate profile via AWS Console

hantonel:
We noticed the steps provided here for the creation of a Fargate profile via AWS Console are not correct, they directly point to "To create an AWS Fargate pod execution role with the AWS Management Console", instead the steps to create the Fargate profile.

Here we can see the instruction to create the Fargate profile with eksctl, however in the tab for "AWS Management Console", the steps are for the creation of the AWS Fargate pod execution role:

--- Getting started with AWS Fargate using Amazon EKS - Create a Fargate profile for your cluster - https://docs.aws.amazon.com/eks/latest/userguide/fargate-getting-started.html#fargate-gs-create-profile
--- https://github.com/awsdocs/amazon-eks-user-guide/blob/master/doc_source/fargate-getting-started.md#create-a-fargate-profile-for-your-cluster

Here is the only place I found properly detailed how to create the Fargate Profile via AWS Console besides eksctl:
--- AWS Fargate profile - Creating a Fargate profile - https://docs.aws.amazon.com/eks/latest/userguide/fargate-profile.html#create-fargate-profile

Could we review this to avoid any confusions?

### DIFF
--- a/doc_source/fargate-getting-started.md
+++ b/doc_source/fargate-getting-started.md
@@ -77,25 +77,34 @@ Create your Fargate profile with the following `eksctl` command, replacing the *
 ```
 eksctl create fargateprofile --cluster <cluster_name> --name <fargate_profile_name> --namespace <kubernetes_namespace> --labels <key=value>
 ```
-
 ------
-#### [ AWS Management Console ]
+#### [ AWS Management Console ]<a name="create-fargate-profile-console"></a>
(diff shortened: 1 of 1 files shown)


jimdial-aws:
Thanks for catching this and bringing it to our attention.

PR #336, opened 2021-03-17
Add note on system: prefix

mikestef9:
*Issue #, if available:*

*Description of changes:*


jimdial-aws:
I've added a note or text to both the eksctl and console sections about not using `system:`, or any portion of that string, for `groupsPrefix` or `usernamePrefix` in both the `eksctl `and console sections, rather than adding this implementation detail to the considerations.

PR #338, opened 2021-03-22
Grammar: added "or"

pychheda:
between AWS Management and Console AWS CloudFormation, understanding that these are two alternative tools to create roles.

*Issue #, if available:*

*Description of changes:*


pychheda:
Also, please remove "the" before AWS Management Console.

jimdial-aws:
Thank you for the submission. I didn't drop "the" however. I'm curious as to why you think it should be dropped though.

PR #345, opened 2021-04-01
Added more clarity on restarting the ebsci controller deployments, ra…

hellodk:
…ther than deleting the controller pods

*Issue #, if available:*

*Description of changes:*

The message in the documentation is somewhat misleading to people, sometimes they may choose to ignore deleting the pods, hence instead of deleting the pods, we've changed it to restart of the ebs-csi-controller deployment.


pgasca:
I'm sorry it took us awhile to get to this, thank you for the contribution! I've verified and merged it into the source. The public website will have the update later this week.

PR #349, opened 2021-04-12
Fixed scaling group ID that will be given higher priority in the section on prioritizing node groups for autoscaling

VijoyChoyi:
Since the highest priority value wins, node groups matching expression ".*p3-node-group.*" will be tried first and only after that auto scaler will match using ".*p2-node-group.*"

*Issue #, if available:*

*Description of changes:*


pgasca:
Thank you, change verified as incorporated.

PR #351, opened 2021-04-12
Add a note to ensure the min or recommended security group settings are

vissree:
met if using a custom security group in ENIConfig

*Issue #, if available:* 350

*Description of changes:*
Add a note linking https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html to ensure that the min or recommended security group settings are met when using a custom security group in the ENIConfig instead of node group security group. 


jimdial-aws:
Thank you for your submission and for using Amazon EKS.

PR #354, opened 2021-04-30
Update create-managed-node-group.md

lboix:
Adding a tip for the usecase where you are describing your whole EKS cluster in just one single file

*Issue #, if available:*

*Description of changes:*


pgasca:
Thank you for the suggestion. I ran it by a few people and we think it may risk confusing some beginner users.

PR #358, opened 2021-05-07
Adding eksbuild versions to update-cluster doc

haouc:
*Issue #, if available:*

*Description of changes:*
- We want to provide the eks build version number to customers, in case that we are releasing new patches. Then customers can be aware of which image can be applied to their cluster addons.
- Bump kube-proxy latest version to 1.19.8 since we have released the new patch version (1.19.8) for EKS as well as kube-proxy.


obrienjason:
Table contents removed.

PR #359, opened 2021-05-12
Bump supported NodeJS AWS SDK version

hardillb:
Following on from https://github.com/aws/aws-sdk-js-v3/issues/2368 it looks like the correct version of the NodeJS AWS SDK should be v3.15 not v2.521.0

*Issue #, if available:*
https://github.com/aws/aws-sdk-js-v3/issues/2368

*Description of changes:*
Following on from https://github.com/aws/aws-sdk-js-v3/issues/2368 it looks like the correct version of the NodeJS AWS SDK should be v3.15 not v2.521.0


ajredniwja:
Hi there, the V2 and V3 are two different SDKs, my only concern it, it might be possible that things work in a specify v2 version as well which needs to be indicated here with `3.15`

hardillb:
V2 and V3 may well be very different SDKs but basic logic implies that 3.x is greater than 2.x so implicitly suggests that 3.x contains all the 2.x contains and more.

If that is not the case, then the docs should explicitly list which version of V2 and which version of V3 is required.

ajredniwja:
3.x does contain more but there are customers who still use 2.x, what I am trying to come at is, if there is a version in 2.x SDK, that should be mentioned as well.

tyrken:
I guess list both v2 and v3 Node SDK versions, as has been done with Java?
```
+ Java \(Version 2\) — [2\.10\.11](https://github.com/aws/aws-sdk-java-v2/releases/tag/2.10.11)
+ Java — [1\.11\.704](https://github.com/aws/aws-sdk-java/releases/tag/1.11.704)
```

hardillb:
OK, so it now lists both v2 and v3 versions.

jimdial-aws:
Thanks for your submission, but I'm going to remove the 2.x version since the version we had has issues and it is not known if there is a 2.x version that does not have issues.

PR #360, opened 2021-05-13
Update security-groups-for-pods.md for revising the sentence regarding security group

0xlen:
*Issue #, if available:*

I think the statement of documentation is mislead. Allowing port 53 for TCP/UDP of cluster security group is not general applicable for affecting the CoreDNS resolution in every environment, unless the security group policy is associating with cluster security group by referring to the following the blog post:

> The first security group we want to apply is the EKS cluster security group, which enables the matched pods launched onto branch network interfaces to communicate with other pods in the cluster such as CoreDNS

- https://aws.amazon.com/tw/blogs/containers/introducing-security-groups-for-pods/

In the EKS workshop, it is associating the Node Group security group accordingly instead of using cluster security group:

- https://www.eksworkshop.com/beginner/115_sg-per-pod/10_secgroup/

*Description of changes:*

So I would like to propose the change by using `Node Group security group`. It would be great if the doc can revise the sentence to make sure it can meet most of scenarios to allow CoreDNS resolution when setting the security group for Pod.


jimdial-aws:
Closing due to a merge conflict, but I've updated the wording to be more specific, rather than referring to security groups names which are more generic and may not exist.

PR #362, opened 2021-05-14
docs update for aws load balancer controller v2.2.0

M00nF1sh:
Documentation updates for aws load balancer controller v2.2.0


jimdial-aws:
Thanks for the submission!

PR #365, opened 2021-05-16
IRSA for CNI while using Windows worker node

Aelshafei:
The steps are lacking any reference to the requirement of configuring IRSA with VPC resource controller in order for this approach to work as expected when detaching the policy "AmazonEKS_CNI_Policy" from the IAM role of the worker nodes

*Issue #, if available:*

*Description of changes:*
added some notes in the relative sections


jimdial-aws:
You no longer need to manage the VPC resource controller in your dataplane, as you can now enable it in the control plane. Please see [Windows support](https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html) for details.

PR #367, opened 2021-05-20
Account ID variable renamed

aka-cafu:
*Issue #, if available:*

*Description of changes:*
Hello,
The Account ID variable exported at step 1 is different from the variable is used to create the trust policy (on step 3).

When running the command to generate the JSON file, the account id field is empty.

Example:


jimdial-aws:
Thank your for your submission!

PR #377, opened 2021-05-28
Fix URLs to avoid 404

netflash:
*Description of changes:*

The given URLs do not work. I just fixed them to avoid 404. Didn't change anything else.


obrienjason:
This content is fixed by PR #378.

PR #385, opened 2021-06-02
`AmazonProvidedDNS`  in the DHCP options set behavior

TheRealGoku:
As per my testing, AmazonProvidedDNS in the DHCP options set is only required when using both "Endpoint public access" as well as "Endpoint private access" as true. In case of scenario where "Endpoint public access" as false and "Endpoint private access" as true, users could use their own dns servers as the customers public dns resolver will essentially yield private ip's that belongs to the range of the EKS cluster subnet.

`AmazonProvidedDNS` as a must statement contradicts as it stops end user to use their private infra networking architecture


TheRealGoku:
Ignore this pull request,

PR #392, opened 2021-06-15
Fix user data example in launch-templates.md

fincd-aws:
*Issue #, if available:*

I get a fatal error that halts `cloud-init` before the bootstrap.sh is run: `/var/lib/cloud/instance/scripts/part-001: line 16: --==MYBOUNDARY==--: command not found`

1. last line of the example has `--==MYBOUNDARY==--\` which doesn't match the declared boundary, because of the trailing `\`.
2. `--==BOUNDARY==--` is not recognized by `cloud-init` (see error above), even though it should be valid from the RFC [RFC 1341](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html). Also I am pretty sure this isn't the last boundary in some cases, as EKS is merging in their own section after this (when NOT using a custom AMI). With a Custom AMI the trailing '--' probably works, it isn't needed though, so removing makes the example work both ways.

*Description of changes:*

Remove the `--\` from the end of the example multipart User Data.

### DIFF
--- a/doc_source/launch-templates.md
+++ b/doc_source/launch-templates.md
@@ -74,7 +74,7 @@ You can combine multiple user data blocks together into a single MIME multi\-par
   #!/bin/bash
   echo "Running custom user data script"
   
-  --==MYBOUNDARY==--\
+  --==MYBOUNDARY==
   ```
 
 ## Using a custom AMI<a name="launch-template-custom-ami"></a>


pgasca:
Thank you @fincd-aws for bringing this to our attention. It seems the trailing backslash was previously taken care of by a [later commit](https://github.com/awsdocs/amazon-eks-user-guide/commit/244c33b11d3438bd4dac5f4eb81a5432ef51449d#diff-d5d30df513a11e81607b882cbd34d991702629d9b1af318a5e121d81e97240ad). As far as the BOUNDARY with the trailing '--', we determined that the example works as how it is written currently in the source. I'm closing this pull request since it seems the backslash removal took care of the issue, but feel free to reopen as needed.

PR #403, opened 2021-06-29
pod to other cluster workload communication

georgio-sd:
The cx had a question about allowing a pod with a SG attached communicate with other cluster workload.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/doc_source/security-groups-for-pods.md
+++ b/doc_source/security-groups-for-pods.md
@@ -106,6 +106,7 @@ The trunk network interface is included in the maximum number of network interfa
 The security groups that you specify in the policy must exist\. If they don't exist, then, when you deploy a pod that matches the selector, your pod remains stuck in the creation process\. If you describe the pod, you'll see an error message similar to the following one: `An error occurred (InvalidSecurityGroupID.NotFound) when calling the CreateNetworkInterface operation: The securityGroup ID '<sg-abc123>' does not exist`\.
 The security group must allow inbound communication from the cluster security group \(for `kubelet`\) over any ports you've configured probes for\.
 The security group must allow outbound communication to the cluster security group \(for CoreDNS\) over TCP and UDP port 53\. The cluster security group must also allow inbound TCP and UDP port 53 communication from all security groups associated to pods\.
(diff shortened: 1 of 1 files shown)


doc_source/security-groups-for-pods.md
```diff
@@ -106,6 +106,7 @@ The trunk network interface is included in the maximum number of network interfa
 The security groups that you specify in the policy must exist\. If they don't exist, then, when you deploy a pod that matches the selector, your pod remains stuck in the creation process\. If you describe the pod, you'll see an error message similar to the following one: `An error occurred (InvalidSecurityGroupID.NotFound) when calling the CreateNetworkInterface operation: The securityGroup ID '<sg-abc123>' does not exist`\.
 The security group must allow inbound communication from the cluster security group \(for `kubelet`\) over any ports you've configured probes for\.
 The security group must allow outbound communication to the cluster security group \(for CoreDNS\) over TCP and UDP port 53\. The cluster security group must also allow inbound TCP and UDP port 53 communication from all security groups associated to pods\.
+The security group must allow outbound communication to the cluster security group \(for other cluster's workload the pod with SGs needs to communicate with\) over necessary ports\. The cluster security group must also allow these inbound ports communication from all security groups associated to pods\.
```
nckturner, line 109: I think this needs to be phrased better, perhaps starting with "If the pod needs to communicate with a workload outside the cluster..."

jimdial-aws:
I'm not totally clear on the request, but I think this is what it's about. Please confirm. If pod A wants to communicate with pod B over say, port 80 on the same cluster, and they both have their own security groups, then the security group on pod A needs to allow outbound 80 to the cluster security group, and the cluster security group needs to accept inbound 80 from pod A's security group? Further, the cluster security group has to allow port 80 outbound to pod B, and pod B's security group needs to allow port 80 inbound from the cluster security group? Is that what you want to get across? In this scenario then, pod A's security group doesn't need port 80 outbound to pod B's security group directly (and pod B doesn't need 80 inbound from pod A's security group directly either)?

jimdial-aws:
This has a merge conflict, but I also recently updated some of the wording that may cover this. Feel free to re-open if you disagree.

PR #410, opened 2021-07-21
Update metrics-server.md

cparik:
*Issue #, if available:* The current step is confusing for new users. 

*Description of changes:*
Added note that the docker pull and push is required only when the worker nodes don't have access to internet to download the metrics server image from public registry.


obrienjason:
This content is being reworked and will consider your contribution. Thank you for the suggestion.

Issue #411, opened 2021-07-23
no mention of --balancing-ignore-label


nbrational:
This relates to issue [kubernetes/autoscaler#3515](https://github.com/kubernetes/autoscaler/issues/3515)

When using `--balance-similar-node-groups=true` to balance scaling across multiple node groups, each in a different AZ, scaling across node group ASGs is not uniform unless all node labels match. 

I created 3 node groups using eksctl (each in a different AZ) and found that a node label `eks.amazonaws.com/sourceLaunchTemplateId` (and possibly `eks.amazonaws.com/sourceLaunchTemplateVersion`) is different across nodes of different node groups. Scaling was not uniform unless I passed `--balancing-ignore-label=eks.amazonaws.com/sourceLaunchTemplateId`. This labels is not covered in the list of [aws ignored labels](https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-1.19.0/cluster-autoscaler/processors/nodegroupset/aws_nodegroups.go) or [BasicIgnoredLabels](https://github.com/kubernetes/autoscaler/blob/cluster-autoscaler-1.19.0/cluster-autoscaler/processors/nodegroupset/compare_nodegroups.go). 

Think it will be useful to include a note in documentation so that user can compare the node labels and pass `--balancing-ignore-label` for each label that needs ignoring.

stevehipwell:
You'll also need https://github.com/kubernetes/autoscaler/issues/3673 to be fixed and the chart to be released (CA chart releases are a manual process after the PR has been merged and the chart has been built).

stevehipwell:
I've opened https://github.com/kubernetes/autoscaler/issues/4228 to request better defaults and documentation around the ignore labels.

pgasca:
I've handled incorporating this information as part of incorporating content for PR. I'm waiting on an engineer to do full confirmation on my summary. But feel free to re-open if there's a need beyond what will be covered with PR: "When setting `--balance-similar-node-groups` to true, make sure that the node groups you want the Cluster Autoscaler to balance have matching labels (except for automatically added zone labels). You can pass a `--balancing-ignore-label` flag to nodes with different labels to balance them regardless, but this should only be done as needed."

PR #412, opened 2021-07-23
mention of `--balancing-ignore-label` in documentation

nbrational:
*Issue #, if available:* https://github.com/awsdocs/amazon-eks-user-guide/issues/411
*Description of changes:*
This relates to issue [kubernetes/autoscaler#3515](https://github.com/kubernetes/autoscaler/issues/3515)

When using --balance-similar-node-groups=true to balance scaling across multiple node groups, each in a different AZ, scaling across node group ASGs is not uniform unless all node labels match.

I created 3 node groups using eksctl (each in a different AZ) and found that a node label eks.amazonaws.com/sourceLaunchTemplateId (and possibly eks.amazonaws.com/sourceLaunchTemplateVersion) is different across nodes of different node groups. Scaling was not uniform unless I passed --balancing-ignore-label=eks.amazonaws.com/sourceLaunchTemplateId. This labels is not covered in the list of aws ignored labels or BasicIgnoredLabels.

Think it will be useful to include a note in documentation so that user can compare the node labels and pass --balancing-ignore-label for each label that needs ignoring.

### DIFF
--- a/doc_source/cluster-autoscaler.md
+++ b/doc_source/cluster-autoscaler.md
@@ -226,7 +226,13 @@ Review the following considerations to optimize your Cluster Autoscaler deployme
 The Cluster Autoscaler can be configured to include any additional features of your nodes\. These features can include Amazon EBS volumes attached to nodes, Amazon EC2 instance types of nodes, or GPU accelerators\. 
 
 **Scope node groups across more than one Availability Zone**  
-We recommend that you configure multiple node groups, scope each group to a single Availability Zone, and enable the `--balance-similar-node-groups` feature\. If you only create one node group, scope that node group to span over more than one Availability Zone\.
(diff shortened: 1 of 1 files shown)


stevehipwell:
@nbrational you might also want to fix the ASG tag definition for taints (see https://github.com/kubernetes/autoscaler/issues/4189) while you're making these changes?

stevehipwell:
@nbrational you might also want to fix the ASG tag definition for taints (see https://github.com/kubernetes/autoscaler/issues/4189) while you're making these changes?

pgasca:
I've incorporated the PR with a few changes:
"When setting --balance-similar-node-groups to true, make sure that the node groups you want the Cluster Autoscaler to balance have matching labels (except for automatically added zone labels). You can pass a --balancing-ignore-label flag to nodes with different labels to balance them regardless, but this should only be done as needed."

If anything else is needed, feel free to submit a new PR.

PR #413, opened 2021-07-23
Fix memory limit

georgio-sd:
CA has a known bug https://github.com/kubernetes/autoscaler/issues/3506
The container consumes more memory than it is limited to.

This fix will prevent issues with OOMKill errors with cluster-autoscaler container

*Issue #, if available:*

https://github.com/kubernetes/autoscaler/issues/3506

*Description of changes:*

### DIFF
--- a/doc_source/cluster-autoscaler.md
+++ b/doc_source/cluster-autoscaler.md
@@ -152,7 +152,14 @@ Complete the following steps to deploy the Cluster Autoscaler\. We recommend tha
      -n kube-system \
      -p '{"spec":{"template":{"metadata":{"annotations":{"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"}}}}}'
    ```
-
+1. Patch the container memory limit to avoid `OOMKill` error in the future\.
+  
+   ```
+   kubectl patch deployment cluster-autoscaler \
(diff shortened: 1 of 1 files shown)


doc_source/cluster-autoscaler.md
```diff
@@ -152,7 +152,14 @@ Complete the following steps to deploy the Cluster Autoscaler\. We recommend tha
      -n kube-system \
      -p '{"spec":{"template":{"metadata":{"annotations":{"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"}}}}}'
    ```
-
+1. Patch the container memory limit to avoid `OOMKill` error in the future\.
+  
+   ```
+   kubectl patch deployment cluster-autoscaler \
+   -n kube-system \
+   -p '{"spec":{"template":{"spec":{"containers":[{"name":"cluster-autoscaler","resources":{"limits":{"memory":"1000Mi"}}}]}}}}'
```
nckturner, line 160: This should probably be left to configure for the customer, with perhaps a suggestion that makes it obvious it is not a 1-size fits all.

obrienjason:
Page content has changed, with memory limits discussed more generally. It's up to the user to determine what's best for their environment.

PR #432, opened 2021-08-10
I fixed wrong parameters in the helm command

georgio-sd:
The original parameters are incorrect:
--------------------
--set serviceAccount.controller.create=true \
--set serviceAccount.controller.name=ebs-csi-controller-sa
--------------------

I checked the helm chart:
--------------------
[root@ip-172-31-4-186 eks-plugins]# helm show all aws-ebs-csi-driver/aws-ebs-csi-driver
...
controller:
...
  serviceAccount:
...
--------------------

They should be:
--------------------
      --set controller.serviceAccount.create=false \
      --set controller.serviceAccount.name=ebs-csi-controller-sa
--------------------

Moreover, we need `false` value here `controller.serviceAccount.create`, since we have created the serviceaccount via eksctl.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/doc_source/ebs-csi.md
+++ b/doc_source/ebs-csi.md
@@ -153,8 +153,8 @@ For detailed descriptions of all the available parameters and complete examples 
       --set image.repository=602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/aws-ebs-csi-driver \
       --set enableVolumeResizing=true \
       --set enableVolumeSnapshot=true \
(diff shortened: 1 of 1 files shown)


pgasca:
Thank you @georgio-sd, this was fixed and elaborated upon as part of another pull request:
https://github.com/awsdocs/amazon-eks-user-guide/pull/451

PR #435, opened 2021-08-16
update curl location for linux

strongjz:
Fix typo on curl location


jimdial-aws:
Thanks for your submission.

PR #436, opened 2021-08-16
Update create-service-account-iam-policy-and-role.md

awstown:
Extra colon in s3 arn example


obrienjason:
Thanks for the contribution!

PR #444, opened 2021-08-26
Improve 'kubectl' cmd call

mtilson:
*Issue #, if available:*
N/A

*Description of changes:*
To remove `kubectl` complaints on deprecated invocation mode and defaulted container, like:
```
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
Defaulted container "aws-node" out of: aws-node, aws-vpc-cni-init (init)
```


jimdial-aws:

PR #445, opened 2021-08-30
Fix edit configmap command

eddiesimeon:
*Issue #, if available:*

Fixed kubectl edit configmap command by removing `-o yaml`


jimdial-aws:
The command works with `-o yaml`, and it's a valid switch, so the command isn't broken with it. As a result, I'm closing this one.

PR #449, opened 2021-09-03
Flag --kubelet-extra-args not needed for containerd when using self managed nodes

damerakd:
There is no need to specify --kubelet-extra-args when using self managed nodes as --container-runtime by itself a independent flag [1] and should not be specified inside --kubelet-extra-args

https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L29


pgasca:
Thank you @damerakd, I found another instance elsewhere as well where I deleted `--kubelet-extra-args`:
```
/etc/eks/bootstrap.sh <my-cluster> \
 --kubelet-extra-args --container-runtime containerd
```

I'll merge it back to GitHub.

PR #450, opened 2021-09-06
Delete labels key from sample ConfigMap

mikutas:
because there is no labels


pgasca:
Thank you!

PR #451, opened 2021-09-06
Breaking change - Regression on Helm Chart values based on user preferred method of installation

TheRealGoku:
This [commit](https://github.com/awsdocs/amazon-eks-user-guide/commit/db7b94af6558dcff6f60950eeb75d0001b786205#diff-e7809e03680f9563e42df4cd5c548f36f2e4265c7aa408c48b70916e5ab9d8b9) helped to solve the way values are referenced in the [values.yaml](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml) file of Helm Chart. However the commit also set ```controller.serviceAccount.create=false``` hence breaking sequence of logical steps that end user follows.

Before committing this pull request:

Steps that end user may follow:
1. Step 1A and 1B to create policy
2. Step 2, I have chosen AWS CLI method [hence no SA getting created]
3. Step 3, I have chosen Helm chart method, which currently has ```controller.serviceAccount.create=false``` hence service account will not get created however the csi controller is referencing this service account. Hence in overall installation perspective ```ebs-csi-node``` DaemonSet pods will get created but ```ebs-csi-controller``` will not get created with error as Service Account not found.

### DIFF
--- a/doc_source/ebs-csi.md
+++ b/doc_source/ebs-csi.md
@@ -141,6 +141,7 @@ For detailed descriptions of all the available parameters and complete examples 
 
    1. Install a release of the driver using the Helm chart\. If your cluster isn't in the `us-west-2` Region, then change `602401143452.dkr.ecr.us-west-2.amazonaws.com` to your Region's [container image address](add-ons-images.md)\.
 
+      If you have used eksctl in Step 2
       ```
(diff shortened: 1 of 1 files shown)


pgasca:
Thank you @TheRealGoku!

PR #458, opened 2021-09-16
Fix kubectl 1.19 download url

vikram-katkar-discovery:


obrienjason:
Thanks for the feedback!

PR #428, opened 2021-08-06
update usage instruction of aws-ebs-csi-driver

andyli:
* controller serviceAccount config moved
 * need to annotate serviceAccount with iam role
 * no enableVolumeResizing and enableVolumeSnapshot
 * default image repo is fine

Re. https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/README.md#upgrading-from-version-1x-to-2x-of-the-helm-chart

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #463, opened 2021-10-01
Remove obsolete enableVolumeResizing and enableVolumeSnapshot ebs csi helm values

wongma7:
*Issue #, if available:* N/A

*Description of changes:* These values were removed from the latest helm chart. The volume resizing and volume snapshot features are enabled by default in both install methods (helm and kustomize/yaml) so setting them isn't needed anymore as it doesn't do anything. Ref: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/release-1.3/charts/aws-ebs-csi-driver/values.yaml

subset of https://github.com/awsdocs/amazon-eks-user-guide/pull/428


pgasca:
Thank you for the PR and the reference!

PR #472, opened 2021-10-28
Added requests and limits secton for EKS Fargate

rsumukha:
Modification of fargate-pod-config to include requests and limits section for Fargate pods.


pgasca:
Merged PR after rewriting it in source and getting it reviewed, will now push periodic update to GitHub for the edited version.

PR #473, opened 2021-11-02
fix wrong configuration in `cloudwatch_logs`

myishay:
*Description of changes:*
With the configuration provided in this page, the logs that are being sent to cloudwatch look like this:
```json
{"log": "{\"data\":\"100 0.5 true This is example\", \"key1\":\"value1\", \"key2\":\"value2\"}"}
```
whereas for properly querying the logs insights, the desired logs format should be:
```json
{"data":"100 0.5 true This is example", "key1":"value1", "key2":"value2"}
```

My changes for the example ConfigMap are preforming theses changes.


myishay:
@pgasca can you please have a look?

nckturner:
I like this change -- especially for querying cloudwatch logs with insights queries.  That being said, what sets the log key to "log"?  Does this work for all configurations, or in what circumstances wouldn't this work?  Do we need a little explanation as to what those fields actually do in these docs (I don't think I fully understand them).

myishay:
> I like this change -- especially for querying cloudwatch logs with insights queries. That being said, what sets the log key to "log"? Does this work for all configurations, or in what circumstances wouldn't this work? Do we need a little explanation as to what those fields actually do in these docs (I don't think I fully understand them).

As far as I understand, FluentBit's data pipelines in most cases are built  as follows:  `[INPUT]` => `[PARSER]`/`[FILTER]` => `[OUTPUT]`

In this specific case, using `aws-logging-cloudwatch-configmap.yaml`, the given `[FILTER]` is using a `[PARSER]` of the type `Regex`.
The given input at the beginning of the pipeline looks something like this:
```
2021-11-01T11:06:18.394065261Z stdout F {"data":"100 0.5 true This is example", "key1":"value1", "key2":"value2"}
```
The `[PARSER]` takes the given regex `^(?<time>[^ ]+) (?<stream>stdout|stderr) (?<logtag>P|F) (?<log>.*)$` and groups the log into 4 groups: `time`, `stream`, `logtag` and `log`.
All these regex groups are being sent from `[PARSER]` back to the `[FILTER]` level as keys in a JSON object, which turns out to be something like this:
```json
{"log":"{\"data\":\"100 0.5 true This is example\", \"key1\":\"value1\", \"key2\":\"value2\"}","time":"2021-11-01T11:06:18.394065261Z","stream":"stdout","logtag":"F"}
```

When configuring `Reserve_Data On` and `Preserve_Key On` under `[FILTER]` stage, the original data  is then being forwarded (and not the parsed data) under `log` key, so there is basically no meaning for the aggregated data in the `[PARSER]` stage.
So, without the changes that I have made, the output from `[FILTER]` stage would be this:
```json
{"log":"2021-11-01T11:06:18.394065261Z stdout F {\"data\":\"100 0.5 true This is example\", \"key1\":\"value1\", \"key2\":\"value2\"}"}
```
And after the changes I made: 
```json
{"log":"{\"data\":\"100 0.5 true This is example\", \"key1\":\"value1\", \"key2\":\"value2\"}","time":"2021-11-01T11:06:18.394065261Z","stream":"stdout","logtag":"F"}
```

Now, by configuring `log_key log` in the `[OUTPUT]` stage, the part that will be sent into CloudWatchonly is only what's under `log` key, which is the desired content to be passed to CloudWatch. If `log_key log` is not configured, the log line that is being sent to CloudWatch is the whole log string as-is (so it's not possible to query with insights queries, since the interesting data to query is parsed as a string and not as JSON.

My EKS Fargate cluster is pretty small and straight-forward (no big tweaks or changes, just followed AWS' instructions to get everything working), so I believe this should be the default configuration for others too.

I had to investigate and invest some time in order to make this configuration work, and it would definitely help me if this config was correctly written in the docs so I could just use it as-is.

nckturner:
Thanks for the detailed explanation!  LGTM. Let's merge after an ok from the fargate team.

myishay:
@nckturner Thank you :)

@pgasca any news on that?

pgasca:
Got approval, thank you @myishay for your great contribution!

PR #477, opened 2021-11-30
update the wording to be less confusing

derom:
By reading this documentation, it's easy to be confused by the word "line" and understand like "add exactly one line from the following block of yaml".


jimdial-aws:
Thank you for your submission and for using Amazon EKS!

PR #485, opened 2021-12-23
fix requirement for ELB VPC Endpoint.

hiraken-w:
*Issue #, if available:*

The requirement for ELB VPC endpoints when creating a Private cluster are not accurate.
Some users are using ALB without the AWS Load Balancer Controller.
And there is no description of NLB.

*Description of changes:*

I changed "Application Load Balancers" to "AWS Load Balancer Controller".


jimdial-aws:
This topic has been changed around a bit, but I have added the endpoint for ALB and NLB.

PR #490, opened 2022-01-18
Added note to use separate IAM roles for the windows and linux nodes

cparik:
*Issue #, if available:*

*Description of changes:*
We have seen users don't use separate roles for their Windows and Linux nodes. 
This causes issues with DNS name resolution similar to [this](https://repost.aws/questions/QUsFW_rlmuRvKpVEs0KMuYoA).
Added a note to not use the same role for both Windows and Linux nodes in the important section before the aws-auth config map. 


pgasca:
Thank you @cparik, I've incorporated the change into source documentation and did a synch back to GitHub. The public website will reflect the change during the next build update.

PR #491, opened 2022-01-20
Update windows support docs for instances suppported

Shreya027:
*Issue #, if available:*
Internal ticket

*Description of changes:*
Added M6a.x as instance not supported for Windows workloads.


obrienjason:
Thanks for the input!

PR #497, opened 2022-02-01
Update efs-csi.md

dushyant8858:
## [Issue Description]
In some cases where Organization that has strict security policies which DO NOT allow any IAM role/user to authenticate/pull images from AWS owned Private registry account, than the documented steps in "Install the Amazon EFS driver" >> "Manifest" does not work.  

## [Description of changes]
Added two OPTION A. (Use images stored in Private ECR Registry) and B. (Use images stored in Public ECR Registry) under "Install the Amazon EFS driver" >> "Manifest" so that user have a flexibility to use the appropriate Registry to pull EFS Driver images.

### DIFF
--- a/doc_source/efs-csi.md
+++ b/doc_source/efs-csi.md
@@ -181,11 +181,13 @@ This procedure requires Helm V3 or later\. To install or upgrade Helm, see [Usin
 ------
 #### [ Manifest ]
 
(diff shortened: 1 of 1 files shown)


pgasca:
Thank you! Merged for GitHub side, and incorporating to source XML doc side with some tweaks that I will send for review.

PR #501, opened 2022-02-09
Fix syntax of controller service account object creation

ohookins:
There were 4 spaces in the `annotations` section so the actual IAM role annotation was ignored by the API server and stripped out, causing the controller to assume the instance role and not be able to function.

*Issue #, if available:*

*Description of changes:*


jimdial-aws:
Thanks for the submission and for using Amazon EKS.

PR #513, opened 2022-03-01
updated doc

malikdraz:
*Issue #, if available:*

Adding following command under cluster-autoscaler deployment does not work. 

--balance-similar-node-groups
--skip-nodes-with-system-pods=false

*Description of changes:*

Removing these lines works perfectly fine with the cluster autoscaler. I have replicated this issue multiple times, come across the same thing. Made changes to deployment without above lines, it works fine and scale the nodes according to need. Please review and remove these. 


obrienjason:
These settings are to rebalance across availability zones and to enable scale to zero. I'll update the documentation to explain this, but these settings are to ensure available computing and to prevent issues with old or underutilized nodes being left in the cluster.

PR #539, opened 2022-04-08
Update csi-iam-role.md

jvahldick:
In the policy condition there is a prefix "oidc-provider/" which is causing credentials issues when the EBS is being created.
I suggest to remove it from the documentation to avoid confusion.

EKS Version: 1.22
Addon: aws-ebs-csi-driver
Addon version: v1.5.2-eksbuild.1


pgasca:
Thank you!

PR #551, opened 2022-04-21
Enhanced documentation for Cluster Autoscaling

guessi:
*Description of changes:*

- red text `<YOUR CLUSTER NAME>` is not easy to find at the end of line.

Reference Link:
- https://docs.aws.amazon.com/eks/latest/userguide/autoscaling.html#cluster-autoscaler

### DIFF
--- a/doc_source/autoscaling.md
+++ b/doc_source/autoscaling.md
@@ -169,7 +169,7 @@ Complete the following steps to deploy the Cluster Autoscaler\. We recommend tha
      -p '{"spec":{"template":{"metadata":{"annotations":{"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"}}}}}'
    ```
 
-1. Edit the Cluster Autoscaler deployment with the following command\.
+1. Edit the Cluster Autoscaler deployment with the following command and replace *`<YOUR CLUSTER NAME>`* with your cluster name\.
 
    ```
    kubectl -n kube-system edit deployment.apps/cluster-autoscaler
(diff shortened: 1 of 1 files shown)


doc_source/autoscaling.md
```diff
@@ -169,7 +169,7 @@ Complete the following steps to deploy the Cluster Autoscaler\. We recommend tha
      -p '{"spec":{"template":{"metadata":{"annotations":{"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"}}}}}'
    ```
 
-1. Edit the Cluster Autoscaler deployment with the following command\.
+1. Edit the Cluster Autoscaler deployment with the following command and replace *`<YOUR CLUSTER NAME>`* with your cluster name\.
```
guessi, line 172: Although it had mentioned in Step 2, but it would be nice to have it mentioned again here.

doc_source/autoscaling.md
```diff
@@ -189,6 +189,7 @@ Complete the following steps to deploy the Cluster Autoscaler\. We recommend tha
            - --cloud-provider=aws
            - --skip-nodes-with-local-storage=false
            - --expander=least-waste
+           # You will need to replace the <YOUR CLUSTER NAME> at the end of line.
            - --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,k8s.io/cluster-autoscaler/<YOUR CLUSTER NAME>
```
guessi, line 193: Need to scroll to the right to discover `<YOUR CLUSTER NAME>` at the end of line if user's screen does not wide enough when viewing the page https://docs.aws.amazon.com/eks/latest/userguide/autoscaling.html#cluster-autoscaler .

obrienjason:
1. The procedure includes a step to replace the <YOUR_CLUSTER_NAME> value already, so this becomes redundant. As this command does not include the YAML file, this would also confuse the user.
2. This is not convention for the guide.
Thanks for the feedback. We do want to improve the documentation and welcome your help.

PR #553, opened 2022-04-22
Reference to full helm chart install documentation

orsenthil:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/tree/main/helm/aws-load-balancer-controller

This explains all the details a customer should take care when going with helm. It is important to give reference to this from aws docs.


jimdial-aws:
I didn't add the sentence. Instead, I added separate Helm and Manifest sections for the removal of the ALB ingress controller or 0.1.0 of the LBC chart (instructions from the GitHub page), so I think that should address it.

Thank you for letting us know.

PR #559, opened 2022-05-02
Update security-groups-for-pods.md

### LINKED ISSUE #557: Improve terminationGracePeriodSeconds guidance in SGP considerations
In [Security groups for pods / considerations](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html#security-groups-pods-considerations), the guidance regarding `terminationGracePeriodSeconds` is easily misunderstood:

> Pods using security groups must contain `terminationGracePeriodSeconds` in their pod spec. This is because the Amazon EKS VPC CNI plugin queries the API server to retrieve the pod IP address before deleting the pod network on the host. Without this setting, the plugin doesn't remove the pod network on the host.

The problems with this statement are:
1. terminatingGracePeriodSeconds has a default value of 30s; why does it have to be present in the "spec"?
2. The [actual condition](https://github.com/aws/amazon-vpc-cni-k8s/issues/1313#issuecomment-901818609) to be avoided is terminatingGracePeriodSeconds=0

joebowbeer:
*Issue #, if available:*

Fixes #557 

*Description of changes:*

Improve guidance for terminationGracePeriodSeconds


jimdial-aws:
Thank you for the contribution.

joebowbeer:
@abhipth Thanks for update. I'm sorry I didn't see it in time to update the PR:

> Pods using security groups must not have `terminationGracePeriodSeconds` set to `0` in their pod spec. (The default value is 30.) This is because the Amazon EKS VPC CNI plugin [prior to v1.10.3] queries the API server to retrieve the pod IP address before deleting the pod network on the host. Without this setting, the plugin doesn't remove the pod network on the host.

In time the entire paragraph can be deleted.

jimdial-aws:
No worries, I'll add the text about the version. I'm also going to modify that sentence altogether, after speaking with the engineers. Thanks again.

PR #572, opened 2022-06-20
Removed the word "bi-directionally"

dumlutimuralp:
1. Word "bi-directionally" implies that resources in the internet can initiate communicate to pods which is not true.
2. Fourth bullet in this doc also makes an assumption that pod in private subnet to an external resource (other VPC, or on-prem or internet) would not work unless the ExternalSNAT is changed from its default, but the communication actually works. So it is worth explicitly mentioning what is the actual reason  to turn on ExternalSNAT=true for this specific case. Because : 
     - For pod -> internet traffic Pod IP gets double SNATed (first by CNI then in IGW layer) and pods are able to access internet.
     - For pod -> on-prem (or pod to a peered VPC) traffic gets SNATed by CNI and then assuming that routing tables are correctly configured once again the traffic pods can reach external resources in on-prem or another VPC.

*Issue #, if available:*

*Description of changes:*


jimdial-aws:
Closing this due to a merge conflict, but I've removed bidirectionally, as you submitted. As for the rest of your text, I made some additional changes to the text that, while it isn't the same as yours, addresses your concerns I think. If not, feel free to re-open the PR. Thanks for your contribution and for using Amazon EKS.

PR #590, opened 2022-08-07
Update Create Cluster Part with an eksctl known issue.

yws-ss:
*Description of changes:*

Update note to avoid unexpected behavior when user want to use eksctl to manage their cluster in the future. 

https://github.com/weaveworks/eksctl/issues/2943


jimdial-aws:
I've incorporated the change, with a bit different text. It will publish later today. Thanks for the contribution.

PR #591, opened 2022-08-10
Update windows-support.md

jiechen0826:
*Issue #, if available:*

*Description of changes:*
Updates the number of back-end pods supported behind a single load balancer from 64 to 1024.


pgasca:
Also found an instance of the 64 limit in load balancing section which I'm updating in source XML and will push to GitHub/website.

jiechen0826:
> 
Awesome, thank you for spotting it!

PR #600, opened 2022-09-09
Fix explanation about autoscaling/v2beta2 in 1.23

mikutas:
Kubernetes 1.23 still supports `apiVersion: autoscaling/v2beta2`. It's just deprecated.
It will be unavailable in 1.26.


jimdial-aws:

PR #602, opened 2022-09-09
Add ebs-csi-controller-sa service account annotation step

luarx:
*Description of changes:*

The annotation step was missing


pgasca:
Thank you @luarx for pointing out this was still missing!

PR #617, opened 2022-10-25
Update update-managed-node-group.md

xfc:
I have to replace No_Schedule with NoSchedule for "taints" to work. The format like "NoSchedule" is same as what's mentioned in Kubenetes documentation.

*Issue #, if available:*

*Description of changes:*


pgasca:
@xfc , while you are correct about the labels in the Console being different, the labels seem to be correct when using `aws eks create-nodegroup` based on my testing of the current AWS CLI procedure. Trying to use `"effect": "NoSchedule"` results in `An error occurred (InvalidParameterException) when calling the CreateNodegroup operation: Invalid taint effect: NoSchedule, must be one of these [NO_SCHEDULE, NO_EXECUTE, PREFER_NO_SCHEDULE]`. Can you give more details on where you had to use `NoSchedule` as the format?

PR #619, opened 2022-10-26
Add some notes on verifying the IAM role

jdn5126:
*Issue #, if available:*
N/A

*Description of changes:*
Customers were occasionally getting confused when creating an IAM role with the `eksctl` command. The most common issue was that a role with that name already existed and so the `eksctl` command did not do anything. This change adds a verification step after the role is created. It also recommends using the cluster name in the role name.


jimdial-aws:
Thanks for the contribution. I'm going to pull the optional steps out into a new step for the entire procedure, so the role can be verified whether it was created with eksctl or the AWS CLI.

PR #621, opened 2022-11-03
Add missing annotation step and pod restart

UrsLange:
The AWS Management Console case is missing the sa annotation step entirely. It is therefore added. After the annotation is made, for it to take effect, the ebs-csi-controller pods need to be restarted. This step is added for AWS Management Console and AWS CLI.


pgasca:
Thank you @UrsLange!

PR #631, opened 2022-11-17
Removes "AWS Management Console" section from launching self-managed …

mmerkes:
…nodes

*Issue #, if available:*

N/A

*Description of changes:*
The [AWS Management Console section](https://docs.aws.amazon.com/eks/latest/userguide/launch-workers.html) references a launch template with a 1.17 nodegroup, which is long deprecated. If customer feedback/comments hasn't motivated us to change this already, we can safely assume that customers don't depend on this and can remove the burden of maintaining it altogether.


pgasca:
Instead of deleting this section, template was updated since it was also used for Outposts.

PR #632, opened 2022-11-20
Added the missing `"` for the `--query` param

htquach:
*Issue #, if available:*

*Description of changes:*
Without the double quotes, the command failed with: no matches found: Stacks[].StackStatus


jimdial-aws:
Thanks for your feedback. In a Bash shell, the command works without error. You may have needed to make an adjustment if you're using a different shell. The prerequisites recommend a Bash shell, but see https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-parameters-quoting-strings.html for adjustments that may need to be made for other shells.

PR #633, opened 2022-11-21
Fixed commands syntax errors

htquach:
Wrapped the environment name with {} when not surrounded by spaces. Specified `--region $region_code` in the AWS CLI commands where they failed if it is not specified.


jimdial-aws:
Thanks for your contribution. You don't need to specify a region code for those commands unless your cluster is in a different region than the default region that you have specified in your AWS [CLI configuration profile](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-parameters-quoting-strings.html). We generally don't specify the --region option with commands for simplicity in the documentation. This is also why the account ID isn't wrapped with {}. It doesn't have spaces around it, so {} aren't used to simplify the documentation.

PR #636, opened 2022-11-25
docs: Mention Node sdk v2 and .NET caveat

philnichol:
*Issue #, if available:*

*Description of changes:*
- This mentions that the Node v2 sdk, which is still being updated and maintained will also work with IRSA if the version is new enough.
- This adds a caveat to the .NET SDK line, since you must include the SecurityToken binary or file to make use of IRSA using .NET. More info in this [PR](https://github.com/aws-samples/one-observability-demo/pull/22) and this [PR](https://github.com/aws/aws-dotnet-extensions-configuration/pull/73/files)

It's understandable if you'd rather not include these, but this info would have helped us massively had we known it beforehand, so figured it might help others. Either way, thanks in advance for your review!


jimdial-aws:

PR #641, opened 2022-12-08
Added changes to getting a token, introduced in 1.24

steveannett:
Added changes to getting a token, introduced in 1.24

In version 1.24 of Kubernetes tokens are no longer automatically generated and the command `kubectl create token` was introduced. See https://github.com/kubernetes/kubernetes/pull/107880 for details.

As the tokens are not automatically generated the guide for getting a token to access the Kubernetes dashboard doesn't work for users on 1.24, and they need to use the command `kubectl -n kube-system create token eks-admin`


JazkyAWS:
Kubernetes Dashboard has been deprecated from the EKS UG.

steveannett:
That's disappointing

PR #648, opened 2022-12-19
Updated checksum checking

qutorial:
The sha256sum -c is an easier way to check the sum. It is a mechanism, relieving the user from needing to manually check the checksum.

*Issue #, if available:*

*Description of changes:*


pgasca:
Thank you @qutorial! I moved your suggested alternative command option to Linux section because it's not available on Mac by default. It seems to be available in CloudShell by default though and out of the box for some other Linux distributions. I also rewrote options for clarity.

PR #663, opened 2023-01-10
Adding note for the fixed value of maxPod enforcement for managed nodegroup

0xlen:
*Issue #, if available:*

I noticed the `maxPod` change when trying to launch the managed nodegroup by using `m5a.4xlarge` instance type, which always limits to `110`, but it should be allowed to run more than this number (`234`) based on [eni-max-pods.txt](https://github.com/awslabs/amazon-eks-ami/blob/e2ef710139a671b2a49b459c0b4654cdc7bab9ba/files/eni-max-pods.txt#L301) file. Eventually I found the change below according to the blog post [1], which stated the optimization is based on the testing by EKS team and Kubernetes scalability thresholds:

> The max pods calculator script limits the return value to 110 based on [Kubernetes scalability thresholds](https://github.com/kubernetes/community/blob/master/sig-scalability/configs-and-limits/thresholds.md) and recommended settings. If your instance type has greater than 30 vCPUs, this limit jumps to 250, a number based on internal EKS scalability team testing.

It gives a clear example to share why the `m5.large` instance should be able to assign `432 IPs` with prefix but only return `110`. In the code snippet [2] I also can see the helper script determine and return the `maxPod` value based on the vCPU. 

In the document [3] regarding enabling prefix assignment, it clearly states this behavior:
> Managed node groups enforces a maximum number on the value of maxPods. For instances with less than 30 vCPUs the maximum number is 110 and for all other instances the maximum number is 250. This maximum number is applied whether prefix delegation is enabled or not.

But it did not mention the historical detail so I was confused when creating my managed nodegroup; hence, I am raising the PR because it would be great to add the notice in the create nodegroup and helper script sections.

Reference:
[1] https://aws.amazon.com/blogs/containers/amazon-vpc-cni-increases-pods-per-node-limits/
[2] https://github.com/awslabs/amazon-eks-ami/blob/e2ef710139a671b2a49b459c0b4654cdc7bab9ba/files/max-pods-calculator.sh#L145-L159
[3] Increase the amount of available IP addresses for your Amazon EC2 nodes - https://docs.aws.amazon.com/eks/latest/userguide/cni-increase-ip-addresses.html

*Description of changes:*
Adding the note to the create nodegroup and the max pods calculator section.


0xlen:
A quick side note is that I noticed the behavior happens when I am spinning up EKS cluster with Kubernetes version `1.23`. But it is not the same behavior when I am running nodegroup with version `1.20` (Control plane and nodegroup both are running consistent Kubernetes version).

- eksctl version: `0.123.0`

```bash
$ eksctl create cluster --name max --instance-types m5a.4xlarge --managed --nodes=2 --region eu-west-1
2023-01-10 16:39:05 [ℹ]  eksctl version 0.123.0
2023-01-10 16:39:05 [ℹ]  using region eu-west-1
...
2023-01-10 16:39:06 [ℹ]  using Kubernetes version 1.23
```

[UserData for 1.23 nodegroup]
```
--//
Content-Type: text/x-shellscript; charset="us-ascii"
#!/bin/bash
set -ex
...
/etc/eks/bootstrap.sh max --kubelet-extra-args '.... --max-pods=110' --b64-cluster-ca $B64_CLUSTER_CA --apiserver-endpoint $API_SERVER_URL --dns-cluster-ip $K8S_CLUSTER_DNS_IP --use-max-pods false

```

PR #664, opened 2023-01-11
Remove duplicate step in EBS CSI driver docs

isaccavalcante:
Remove duplicate step (**5.b** and **6**) from the EBS CSI drivers documentation

image


jimdial-aws:

PR #670, opened 2023-01-24
Show correct latest version of coredns addon for EKS 1.22

dmcnaught:
v1.8.7-eksbuild.3 is not available on EKS 1.22, latest is v1.8.7-eksbuild.1

*Issue #, if available:*

*Description of changes:*


jimdial-aws:
You're correct about the available version for the Amazon EKS type of the add-on. The table however, is also correct, as it lists the latest available container image tag (for the self-managed version). Unfortunately, the available image versions don't always match the available Amazon EKS add-on versions. I've recently updated the documentation to (hopefully) make the differences between the versions more clear.

Thanks for using Amazon EKS.

PR #671, opened 2023-01-24
escaping special character

bniu-amz:
current cmd result in "zsh: no matches found: Stacks[].StackStatus" on mac.

aws cloudformation describe-stacks --stack-name my-eks-custom-networking-vpc --query Stacks\[\].StackStatus --output text  fixes the issue.

*Issue #, if available:*

*Description of changes:*


jimdial-aws:

PR #714, opened 2023-04-12
Update installation instructions to remove third-party installers

yuxiang-zhang:
*Issue #, if available:*

*Description of changes:*
Update installation instructions for eksctl corresponding to https://github.com/weaveworks/eksctl/pull/6376


pgasca:
The intent behind the PR is to make sure things are at parity with the official eksctl installation instructions. To avoid things going out of sync, we try to have only one main source of steps rather than have redundancy spread across several sites. So on top of this change, I'm going to do more with the approach of linking to the official instructions from the EKS User Guide.

PR #716, opened 2023-04-19
KMS key ID changed to ARN for the IAM Policy

andrzej-mega:
*Description of changes:*
IAM policy expects an ARN for the resource property. Key ID changed to Key ARN in the documentation.


pgasca:
Thank you @andrzej-mega! I've also made the same suggested change to other alternative procedures and added in-line links to references for more information.

PR #729, opened 2023-05-17
Update managing-kube-proxy.md

orsenthil:
Only minimal type images are shipped with EKS 1.25 on wards.

*Issue #, if available:*

Address an internal issue raised wherein they noticed "default type" of EKS 1.25 had  different iptables version than "default type" of kube-proxy in EKS 1.24.

This explanation will clarify the difference.

*Description of changes:*

* Only minimal type images are shipped with EKS 1.25 on wards.


orsenthil:
@jimdial-aws  - Please take a look.

orsenthil:
Hi @jimdial-aws , @JazkyAWS , when will this be reflected in the published documentation?

jimdial-aws:
It was published after the PR was merged, just not exactly as you wrote it. See the table and the first bullet in the note under the table in https://docs.aws.amazon.com/eks/latest/userguide/managing-kube-proxy.html

PR #737, opened 2023-09-07
Fix command to edit aws-node for CNI network policy doc

oxr463:
*Issue #, if available:*

*Description of changes:*

The command to edit the `aws-node` daemonset was wrong, so I updated it.

### DIFF
--- a/doc_source/cni-network-policy.md
+++ b/doc_source/cni-network-policy.md
@@ -142,7 +142,7 @@ For all other cluster versions, if you upgrade the Amazon EKS optimized Amazon L
         1. Open the `aws-node` `DaemonSet` in your editor\.
 
            ```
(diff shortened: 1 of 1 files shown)


fincd-aws:
Thanks for contributing!

Having `-o yaml` on the end does work for me but I'm pretty sure its a no-op - you always get YAML with `kubectl edit`.

fincd-aws:
Oops!

Hopefully this is the only time this happens, but I merged this into `master`, not `main`.
I merged manually (git cli) again into `main` here: https://github.com/awsdocs/amazon-eks-user-guide/commit/c0ea43e908f6c365fc29dc72fad4f5454c636619 
This way I preserve your original commit ID which is GPG verified.

The docs website will update shortly - the same `kubectl daemonset -n kube-system amazon-vpc-cni` was copied in multiple places so I'm fixing those too.

PR #742, opened 2023-11-14
fix: s3 gateway endpoint

alpkahvecioglu:
*Description of changes:* on the Managed Node Groups page there is a list of VPC endpoints that users must create if they want to use Private Subnets. It seems like there is a mistake in the S3 endpoint.


pgasca:
Thank you, confirmed that EKS guide contradicted the source it was pointing to.

PR #748, opened 2024-03-27
sgp fix:remove p3 instance type support and update link to limits.go

sushrk:
*Issue #, if available:*
N/A
*Description of changes:*
P3 is in the Xen instance family and not a nitro-based instance. ENI trunking does not work reliably on non-Nitro instances. 
Update doc to remove P3 instance type. 

Also updated the link for `limits.go` file as v1.1.4 is outdated. We have a `Notice` section in each release notes which links to the corresponding `limits.go` file, https://github.com/aws/amazon-vpc-resource-controller-k8s/releases.


doc_source/security-groups-for-pods.md
```diff
@@ -6,7 +6,7 @@ Security groups for Pods integrate Amazon EC2 security groups with Kubernetes Po
...
++ Security groups for Pods are supported by most [Nitro\-based](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances) Amazon EC2 instance families, though not by all generations of a family\. For example, the `m5`, `c5`, `r5`, `m6g`, `c6g`, and `r6g` instance family and generations are supported\. No instance types in the `t` family are supported\. For a complete list of supported instance types, see the `limits.go` file linked in `Notice` for the relevant [release](https://github.com/aws/amazon-vpc-resource-controller-k8s/releases) on Github\. Your nodes must be one of the listed instance types that have `IsTrunkingCompatible: true` in that file\.
```
fincd-aws, line 9: Let's keep the link directly to the limits.go and just increment the version to 1.4.5/1.4.6 (1.4.6 is tagged in that repo but not released).
as in: `https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/v1.4.6/pkg/aws/vpc/limits.go`

Linking to other text that has links in it doesn't work out; many folks don't want to follow multiple steps to leave the docs before getting to the info. I can afford the extra step every few weeks to increment this version when you release.

At a later date, I'll revisit doing some "transclude" to copy data in from other GitHub sources - incrementing these versions, converting JSON to lists like this limits.go, pulling info from APIs, etc.

PR #754, opened 2024-05-21
Pod IPv4 and IPv6 assignment update

dumlutimuralp:
Pods are assigned a host-local IPv4 address, this is a specific feature in VPC CNI that must be mentioned.

*Issue #, if available:*
Pods are assigned a host-local IPv4 address, this is a specific feature in VPC CNI that must be mentioned. 

*Description of changes:*
Split the original 12th bullet into two bullets. The ones starting with "Kubernetes Services are..." and "Pods are assigned".


geoffcline:
Thanks! I've staged these changes.

PR #768, opened 2024-06-17
Correct configuration values for updating CoreDNS auto-scaling via CLI

abstrask:
*Description of changes:*

`minReplicas` and `maxReplicas` need to go inside the `autoScaling` block.

Current documented command returns:

```
An error occurred (InvalidParameterException) when calling the UpdateAddon operation: ConfigurationValue provided in request is not supported: Json schema validation failed with error: [$.minReplicas: is not defined in the schema and the schema does not allow additional properties, $.maxReplicas: is not defined in the schema and the schema does not allow additional properties]
```


geoffcline:
great catch! I've staged these changes.

PR #773, opened 2024-07-02
Update s3-csi.md - Prerequisite

Ramneek-kalra:
Updated one more prerequiste which I tested on my own as well.

*Issue #, if available:*
Without EFS CSI Driver installed, S3 CSI Driver won't work and will throw below error:
```
0/2 nodes are available: pod has unbound immediate PersistentVolumeClaims. preemption: 0/2 nodes are available: 2 Preemption is not helpful for scheduling..
```

*Description of changes:*
Addition of a point i.e., "While using S3 CSI Driver, we require EFS CSI Driver as storage won't be provisioned to get PersistentVolume mounted." I have tested this as well today and it's working.

Checkout related Issue: https://github.com/awslabs/mountpoint-s3-csi-driver/issues/185#issuecomment-2097980816

### DIFF
--- a/doc_source/s3-csi.md
+++ b/doc_source/s3-csi.md
@@ -12,6 +12,7 @@ Static provisioning refers to using an existing Amazon S3 bucket that is specifi
 
 **Prerequisites**
 + An existing AWS Identity and Access Management \(IAM\) OpenID Connect \(OIDC\) provider for your cluster\. To determine whether you already have one, or to create one, see [Create an IAM OIDC provider for your cluster](enable-iam-roles-for-service-accounts.md)\.
++ An existing EFS CSI Driver is also required as storage won't be provisioned without that. Check [Related Issue.](https://github.com/awslabs/mountpoint-s3-csi-driver/issues/185#issuecomment-2097980816)
 + Version 2\.12\.3 or later of the AWS CLI installed and configured on your device or AWS CloudShell\.
 + The `kubectl` command line tool is installed on your device or AWS CloudShell\. The version can be the same as or up to one minor version earlier or later than the Kubernetes version of your cluster\. For example, if your cluster version is `1.29`, you can use `kubectl` version `1.28`, `1.29`, or `1.30` with it\. To install or upgrade `kubectl`, see [Installing or updating `kubectl`](install-kubectl.md)\.
 


chrisnegus:
Thanks @Ramneek-kalra. Changes have been staged.

mveitas:
@Ramneek-kalra @chrisnegus Is this change suggesting that the EFS CSI driver is a prerequisites to using the S3 CSI driver?

The documentation [here](https://docs.aws.amazon.com/eks/latest/userguide/s3-csi.html) seems to suggest the EFS driver is not provisioned

dannycjones:
This change shouldn't be required, Mountpoint S3 CSI Driver does not require EFS CSI Driver to be installed.

If somethings not working and installing EFS CSI Driver seems to help, please do [open an issue on the Mountpoint S3 CSI Driver repository](https://github.com/awslabs/mountpoint-s3-csi-driver/issues/new) so the team can investigate. I saw there was already a comment in https://github.com/awslabs/mountpoint-s3-csi-driver/issues/185#issuecomment-2097980816 but we should have a dedicated issue to dig deeper, as we'd need logs to investigate.

towhidul1987:
I did some testing based on a Customer requirement who want to use S3 and EFS CSI Driver parallel in the same Cluster. The below Consideration seems invalid. [Document](https://docs.aws.amazon.com/eks/latest/userguide/s3-csi.html)
"To get PersistentVolume mounted while using the Mountpoint for Amazon S3 CSI driver, we require that the Amazon EFS CSI driver won't be provisioned."

EKS Version: 1.29
S3-CSI-Driver: v1.7.0
EFS-CSI-Driver: v2.0.6

Only S3-CSI Driver Installed in the Cluster: POD is running fine.
S3+EFS CSI both Driver Installed in the Cluster: Both POD using S3 and EFS running Fine.
S3+EFS CSI Driver both Driver Installed in the Cluster: Both POD using S3 and EFS running Fine in the Same WorkerNode as well.

I don't see any relation between S3 and EFS CSI Driver.

pgasca:
Thank you @towhidul1987, I confirmed with several people that this statement was misleading and it has been removed. The main documentation website will have that consideration removed in about 2 hours.

PR #786, opened 2024-08-21
Update quickstart.md

rysloan4:
*Issue #, if available:*

*Description of changes:*
Add step for making sure kubeconfig is correctly configured to interact with the cluster. Otherwise the following helm command will fail

### DIFF
--- a/doc_source/quickstart.md
+++ b/doc_source/quickstart.md
@@ -115,6 +115,11 @@ The AWS Load Balancer Controller \(LBC\) leverages Custom Resource Definitions \
    ```
    helm repo update eks
    ```
+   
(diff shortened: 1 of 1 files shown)


fincd-aws:
Thanks for this! Paul, can you take a look?

pgasca:
The suggested step isn't needed. When you run the command `eksctl create cluster -f cluster-config.yaml` in the earlier step, it also creates a `kubectl` config file in `~/.kube` or adds the new cluster's configuration within an existing config file in `~/.kube`. We want to keep the quick start procedure succinct so I don't think adding that information would be helpful in this case.

PR #795, opened 2024-09-23
Update network-reqs.md

yuwesu:
*Issue #, if available:*

*Description of changes:*

When create EKS cluster, user need to select two subnets in different AZs. If customer only choose 1 subnet or 2 subnets in same AZs, EKS API will get error message:

`InvalidParameterException: Subnets specified must be in at least two different AZs`


fincd-aws:
Hey, thanks for this!

I was sure we already mentioned that somewhere, but I don't see it now, and this is a good spot to put it.

I'll merge and deploy, changes won't be visible for a couple hours.

geoffcline:
Hello! 👋

I've captured your suggested changes in our internal system as we prepare to migrate this repository from Markdown to AsciiDoc format. We'll be offline for about a week during this transition, so I'm closing this PR for housekeeping purposes.

Your contribution will be included when we republish with the new format. Thank you for helping improve our AWS documentation!

PR #798, opened 2024-09-26
Added PDB details to 'Edit a nodegroup' section

pgasca:
Added further details as discussed with internal team regarding PDBs not being considered during an EKS MNG scalingConfig modification. Multiple customers were confused on the fact that EKS MNG do not consider PDBs, so added the **Important** section at the bottom to clarify that changing the scalingConfig of an EKS MNG does not follow the same workflow as updating an EKS MNG.

*Issue #, if available:*

*Description of changes:*


pgasca:
Merged and now making some edits for next refresh pass.

PR #800, opened 2024-09-27
Update AMI config in launch-node-ubuntu.md

carlosbravoa:
*Issue #, if available:*

*Description of changes:*
Removed a dot in the AMI family name which was creating problems and removed the AMI: line, as not only it is not needed but also created an error with the deployment


geoffcline:
Hello! 👋

I've captured your suggested changes in our internal system as we prepare to migrate this repository from Markdown to AsciiDoc format. We'll be offline for about a week during this transition, so I'm closing this PR for housekeeping purposes.

Your contribution will be included when we republish with the new format. Thank you for helping improve our AWS documentation!

PR #805, opened 2024-10-17
Nambinayagan patch 1

nambinayagan:
*Issue #, if available:* Using the `aws eks update-cluster-config` command provided in disable-extended-support and enable-extended-support documentation caused below error:

>> Unknown options: STANDARD, =

*Description of changes:* The error was because the command to enable/disable Extended support had `supportType = string` where the additional spaces before and after "=" was causing issue. So changed it to `supportType=string` which follows the correct shorthand Syntax. 


geoffcline:
fixed, thanks!

PR #812, opened 2024-11-25
Update IAM role policy configuration to use `StringEquals` for Mountpoint for Amazon S3 CSI driver

jiaeenie:
Currently, the documentation recommends that customers use a wildcard match `StringLike`  in their IAM role policy configuration for the CSI Driver EKS add-on. This approach is not considered a security best practice and has caused confusions for some customers (see: https://github.com/awslabs/mountpoint-s3-csi-driver/issues/300 and https://github.com/awslabs/mountpoint-s3-csi-driver/issues/173#issuecomment-2213021838).

In this pull request, we update the documentation to advice using `StringEquals` instead of `StringLike` to improve security and address these concerns.

/cc @unexge @muddyfish @dannycjones 

### DIFF
--- a/latest/ug/storage/s3-csi.adoc
+++ b/latest/ug/storage/s3-csi.adoc
@@ -185,9 +185,9 @@ Add a comma to the end of the previous line, and then add the following line aft
 +
 [source,json,subs="verbatim,attributes,quotes"]
 ----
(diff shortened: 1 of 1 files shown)


unexge (approved):
LGTM

pgasca (approved):

PR #820, opened 2025-01-01
Fix links in troubleshooting.adoc

fincd-aws:
*Issue #, if available:*

*Description of changes:* Fix broken links on this page by converting the type of links.


pgasca (commented):
This link approach is a workaround whenever we can't figure out some syntax error that is elsewhere in the file. I have solved the source of the link rendering issue so this workaround is no longer needed.

PR #821, opened 2025-01-02
Fixed the fomatting of the broken links on the troubleshooting page.

asjarre:
…mat to long format link:

*Issue #, if available:*

*Description of changes:*
Fixed the fomatting of the broken links by changing from the <<>> to longer link format link:


pgasca:
@asjarre Thank you for your suggestion of an external link workaround, but I found an alternative approach that allows keeping the preferred cross-reference syntax. The problem was elsewhere in the file: opening inline code with double backticks instead of single backticks.

The fix isn't live yet on the public website but should be there for next refresh.

PR #824, opened 2025-01-09
add missing step for vpa installation

lsdeva:
Fix for error MountVolume.SetUp failed for volume "tls-certs" : secret "vpa-tls-certs" not found.

https://github.com/kubernetes/autoscaler/issues/4775

*Issue #, if available:*

*Description of changes:*


lsdeva:
If directly invoke` ./hack/vpa-up.sh ` without vpa-tls-certs, vpa-admission-controller pod will stall in ContainerCreating stage.

image

The issue was discussed and the solution was proposed at https://github.com/kubernetes/autoscaler/issues/4775

geoffcline:
thanks for opening this PR 🥳 

I'm researching further

geoffcline:
hey @lsdeva sorry this was forced closed -- doing repo maintenance 

i've made a note of the requested changes and will recreate, apologies

voelzmo:
Hey, VPA maintainer here 👋 
I don't think adding this step should be necessary, this is actually a bug in the vpa 1.2 branch. The fix is on the way and should be released with vpa 1.2.2 soon: https://github.com/kubernetes/autoscaler/pull/7704

If you want to add this as information for users installing vpa 1.2.1 or 1.2.0, you certainly can – but maybe the advice here would rather be to switch to the newer patch release then.

PR #835, opened 2025-01-15
Update launch-node-ubuntu.adoc

carlosbravoa:
Removing `ami: auto-ssm` as it makes "think" eksctl that it is a custom ami, throwing an error (needing `overrideBootstrapCommand`)

``` 
Error: nodeGroups[0].overrideBootstrapCommand is required when using a custom AMI based on Ubuntu2204 (nodeGroups[0].ami) 
```

*Issue #, if available:*

*Description of changes:*


maiconrocha:
I tested, and it works successfully with the change implemented by @carlosbravoa. Thanks for your contribution.

PR #846, opened 2025-01-21
Update efs-csi.adoc

nisi-zx:
There was a typo on the documentation where Step 3 provided a Note stating that Fargate does not require the creation of the EFS CSI Driver where it should listed under Step 2

*Issue #, if available:*

Step 3 provides guidance on creating an EFS Volume and provides note stating that this step is not necessary for Fargate and Fargate does not require installation of EFS CSI Driver -- this should be outlined in Step 2 instead.

*Description of changes:*

Proposed movement of Note to Step 2 instead of Step 3 as it is misleading.


pgasca (approved):
Agreed, this note was placed in the wrong location.

PR #871, opened 2025-02-07
Adds updating-related information for EKS Local Clusters on Outposts

alegermer:
*Description of changes:*

Extends EKS local-clusters on outposts documentation with some clarification regarding the EKS Platform-Version update process and related troubleshooting information. The gist of what is being addressed here is:

* Clearly state that EC2 instances related to the managed Kubernetes control-plane are automatically recycled during platform updates.
* Explicitly call out risks associated with manual termination of these managed instances.
* Provide troubleshooting guidance for customers that may suddenly find their clusters stuck in `UPDATING` state.


alegermer:
Thanks for the improvements @fincd-aws! It looks good to me!

geoffcline:
@fincd-aws is this good to merge?

PR #880, opened 2025-02-13
Update IAM policy requirements for EKS worker nodes

devanshpoplii:
This PR updates the documentation to clarify that both the AmazonEC2ContainerRegistryPullOnly and AmazonEC2ContainerRegistryReadOnly policies are required for EKS worker nodes. 
The documentation previously mentioned only the AmazonEC2ContainerRegistryPullOnly policy,  but during node group creation, the AmazonEC2ContainerRegistryReadOnly policy is also required.

*Issue #, if available:*

*Description of changes:*
- Added AmazonEC2ContainerRegistryReadOnly policy requirement alongside AmazonEC2ContainerRegistryPullOnly.
- Updated instructions to reflect the need for both policies when creating a worker node role.


geoffcline:
Thanks for opening @devanshpoplii 

We've made an internal note about this, and we will push an update based on that.

PR #882, opened 2025-02-15
Minor fixes and simplification of hybrid nodes upgrade instructions

gauravpriyolkar:
*Description of changes:* 
- Simplified steps for marking a node as unschedulable by using cordon instead of adding taints. Also, uncordon step is missing in the in-place upgrade instructions on the live site, but I see it has been added recently to the markdown in git, but not yet published.
- Fixed command for checking status of hybrid nodes after upgrade.
- Fix command to identify all nodes of a particular Kubernetes version in cutover migration upgrade instructions.


csplinter (approved):

chrisnegus (approved):
looks good to me.

PR #907, opened 2025-03-06
fix: broken NetworkPolicy example URLs

andskli:
The eksworkshop.com links are outdated, the content referenced is available at archive.eksworkshop.com

---


fincd-aws (approved):
Verified several of the links.

The EKS Workshop now uses the more complex shopping app. We can consider switching this demo over to it later.

PR #908, opened 2025-03-06
fix: aws eks update-cluster-config example

angonz:
Remove a redundant line in the example of `aws eks update-cluster-config` command to disable default node pools


geoffcline:
thanks!

PR #909, opened 2025-03-07
Remove duplicated role entry en create-node-class.adoc

angonz:
Remove a duplicated reference to key `role` in the node class specification example. I think that the second one is incorrect, as it has an arn as value. Please double check with the actual node class specification.


fincd-aws:
Hey, I changed this the other way, keeping the lower one.
I'll close this PR but many thanks for submitting it!

PR #922, opened 2025-03-14
Add note about kube-proxy 1.21 and lower not compatible with k8s 1.25

david-a-aws:
*Issue #, if available:*
.
*Description of changes:*
Seeing customers experience issues with kube-proxy after their cluster is upgraded to 1.25 (usually automatic due to extend support version deprecation). This causes critical issues for their workloads, as pods that rely on kube-proxy will fail.
Kube-proxy will show errors relating to EndpointSlices, as there is version incompatibility.

I can see the update to use V1 was in [k8s 1.22](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md)
> Kube-proxy uses V1 EndpointSlices. (#103306, @swetharepakula)


fincd-aws:
That is a specific reason why kube-proxy 1.21 doesn't work with k8s 1.25, but in general that scenario is already unsupported.

I'd prefer not to add a specific note, instead emphasize that the allowed skew in the versions of k8s components was 2 back then:
> `kube-proxy` < 1.25 may only be up to two minor versions older than kube-apiserver 

From the K8s docs: https://kubernetes.io/releases/version-skew-policy/

david-a-aws:
> That is a specific reason why kube-proxy 1.21 doesn't work with k8s 1.25, but in general that scenario is already unsupported.
> 
> I'd prefer not to add a specific note, instead emphasize that the allowed skew in the versions of k8s components was 2 back then:
> 
> > `kube-proxy` < 1.25 may only be up to two minor versions older than kube-apiserver
> 
> From the K8s docs: https://kubernetes.io/releases/version-skew-policy/

Supported or not, customers are running into this issue and it is having production impact. I think it should be clearly stated, as it is with the other important notes. Also, it is a cluster-breaking change which impacts any pod that relies on kube-proxy including CoreDNS, so I'd argue it is the most important note to be aware of for this release.

geoffcline:
off topic @david-a-aws but will you associate your github profile with AWS using OpenSourcerer?

david-a-aws:
> off topic @david-a-aws but will you associate your github profile with AWS using OpenSourcerer?

I have done that now :)

geoffcline:
reviewed and decided not to merge

PR #923, opened 2025-03-16
Fixed curl command in lbc-helm.adoc 

asjarre:
*Description of changes:*
The load balancer controller Helm instructions use the incorrect URL with "curl -O". This command will download the full web page:
curl -O https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/install/iam_policy.json

Instead, the raw.githubusercontent URL should be used:
curl -O https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/v2.11.0/docs/install/iam_policy.json

The updated command is the same as the one used in the load balancer controller manifest documentation.


fincd-aws:
Hey, thanks for this!
I will update the docs website in a few hours.

PR #924, opened 2025-03-17
Update windows-support.adoc with access entry considerations 

geoffcline:
add consideration for access entries type


latest/ug/clusters/windows-support.adoc
```diff
@@ -72,7 +78,8 @@ aws iam attach-role-policy \
   --role-name eksClusterRole \
   --policy-arn {arn-aws}iam::aws:policy/AmazonEKSVPCResourceController
 ----
-. Create a file named [.replaceable]`vpc-resource-controller-configmap.yaml` with the following contents.
+. If the VPC CNI *is not* installed as an Amazon EKS Add-on:
```
KlwntSingh: VPC CNI is is always installed for networking in k8s. This command just enable Windows-ipam inside VPC resource controller.

Rephrase the statement to, enable IPAM for Windows nodes in VPC resource controller using following ConfigMap.

latest/ug/clusters/windows-support.adoc
```diff
@@ -29,13 +32,16 @@ Before deploying Windows nodes, be aware of the following considerations.
 * The source for the controller is managed on GitHub. To contribute to, or file issues against the controller, visit the https://github.com/aws/amazon-vpc-resource-controller-k8s[project] on GitHub.
 * When specifying a custom AMI ID for Windows managed node groups, add `eks:kube-proxy-windows` to your {aws} IAM Authenticator configuration map. For more information, see <<mng-ami-id-conditions>>.
 * If preserving your available IPv4 addresses is crucial for your subnet, refer to https://aws.github.io/aws-eks-best-practices/windows/docs/networking/#ip-address-management[EKS Best Practices Guide - Windows Networking IP Address Management] for guidance. 
+* Considerations for EKS Access Entries
+** If you use a different Node IAM Role for Windows instances, EKS will automatically create the required Windows Access Entry. 
+** Access Entries for use with Windows nodes need the type of `EC2_WINDOWS`. For more information, see <<creating-access-entries>>.
```
KlwntSingh, line 38: eg for creating access entry for Windows node
```
aws eks create-access-entry --cluster-name my-cluster --principal-arn arn:aws:iam::111122223333:role/EKS-my-cluster-self-managed-ng-1 --type EC2_Windows
```

latest/ug/clusters/windows-support.adoc
geoffcline: I think it matters if it's installed as a managed add-on or regular, right?

if it's a managed add-on, they need to configure the add-on

if it's regular, they can just update the config

does that make sense?

KlwntSingh: CNI for Windows is different from Linux. Windows CNI is shipped in AMIs. It is not installed via addon.

geoffcline: ah tyty

PR #930, opened 2025-03-19
Fix zsh globbing pattern in aws iam list-attached-role-policies query

mattijons:
zsh interprets `[]` as a globbing pattern:

```
$ aws iam list-attached-role-policies --role-name my-role --query AttachedPolicies[].PolicyArn --output text
$ zsh: no matches found: AttachedPolicies[].PolicyArn
```

*Description of changes:*
Add quotes around `AttachedPolicies[].PolicyArn`


geoffcline:
Thanks!

PR #932, opened 2025-03-20
Fix containerd/kubelet proxy configuration docs for EKS Hybrid nodes

abhay-krishna:
Fix containerd/kubelet proxy configuration docs for EKS Hybrid nodes.


chrisnegus (approved):
lgtm

PR #944, opened 2025-03-26
add summary of the EKS Cluster upgrade process

maiconrocha:
*Issue #, if available:*

*Description of changes:*
Add summary of the EKS Cluster upgrade process


pgasca (approved):
In my rewrites, I removed the "Step" labels because it's a bit confusing that they don't fully line up with the main steps in the full procedure. Later on, the procedure itself or summary could be further rewritten to better align with each other (one summary step per main step section).

PR #964, opened 2025-04-03
be more specific about Node replacement occuring due to AMI drift

tzneal:
*Issue #, if available:*

*Description of changes:*

Small update to call out that by default Nodes will be replaced when new AMIs are released.


geoffcline:
Thanks @tzneal -- I'm not merging this quite yet as I may revise to include this callout in other places.

PR #985, opened 2025-04-16
Update creating-access-entries.adoc

Veronica4036:
We see this error in the console but it's not documented anywhere, hence raising PR for the same. The Kubernetes group name cannot start with "system:".

*Description of changes:* Added documentation about restriction on Kubernetes group names starting with "system:" in access entries. This helps users understand and avoid a common error message shown in the console.
image


geoffcline:
I appreciate you opening this PR

I have decided not to merge it because the same consideration is currently on the page

https://github.com/awsdocs/amazon-eks-user-guide/pull/985/files#diff-20f5ebd16f73ab666bee3c256d4c3f05dfca049a80359106a05b9721ac1cd5e9R29

We have to be extra cautious about adding information to warning boxes. In this case, since the console detects the issue, we shouldn't put it in a warning box.

Thanks again :)

Veronica4036:
@geoffcline May I know where is the consideration mentioned?

geoffcline:
Screenshot 2025-04-17 at 7 40 04 PM

Yes, it's here

https://docs.aws.amazon.com/eks/latest/userguide/creating-access-entries.html

I can take the feedback that the overall page needs revising for clarity.

Veronica4036:
Thats for the usernames and not the groups

PR #998, opened 2025-05-02
Update eks-ami-deprecation-faqs.adoc

terana:
*Issue #, if available:*

*Description of changes:*

- Fix the latest kernel version used on AL2023: it's not 6.10, but 6.12 [1]
- Remove the mention that AL2023 doesn't support graphical desktop: support was added in March release [2]
[1] https://docs.aws.amazon.com/linux/al2023/ug/compare-with-al2-kernel.html
[2] https://docs.aws.amazon.com/linux/al2023/ug/graphical-desktop-al2023.html


fincd-aws:
Hey, thanks for this!

The latest AL2023 AMI from EKS still has kernel 6.1, not 6.10 or 6.12. I know we're testing it though.

Graphical desktop is a good callout, its never gonna be pre-installed on EKS AMIs but you can do it.

EDIT: 6.1.132, from https://github.com/awslabs/amazon-eks-ami/releases/tag/v20250419

geoffcline:
@fincd-aws does this look good now?

PR #1004, opened 2025-05-08
broken documentation of aws-load-balancer-controller after v2.13 release

Ahmad-RW:
*Description of changes:*
Updated docs to reflect new release (v2.13) as the doc right now is broken since the helm chart will install the latest version however step 1 is fixed on installing v2.12 policy

### DIFF
--- a/latest/ug/networking/lbc-helm.adoc
+++ b/latest/ug/networking/lbc-helm.adoc
@@ -43,13 +43,13 @@ Before proceeding with the configuration steps on this page, consider the follow
 
 [#lbc-helm-iam]
 == Step 1: Create IAM Role using `eksctl`
-The following steps refer to the {aws} Load Balancer Controller *v2.12.0* release version. For more information about all releases, see the https://github.com/kubernetes-sigs/aws-load-balancer-controller/releases/[{aws} Load Balancer Controller Release Page] on GitHub.
+The following steps refer to the {aws} Load Balancer Controller *v2.13.0* release version. For more information about all releases, see the https://github.com/kubernetes-sigs/aws-load-balancer-controller/releases/[{aws} Load Balancer Controller Release Page] on GitHub.
 
 . Download an IAM policy for the {aws} Load Balancer Controller that allows it to make calls to {aws} APIs on your behalf. 
 +
 [source,shell,subs="verbatim,attributes"]
 ----
-curl -O https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/v2.12.0/docs/install/iam_policy.json
+curl -O https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/v2.13.0/docs/install/iam_policy.json
 ----
 ** If you are a non-standard {aws} partition, such as a Government or China region, https://github.com/kubernetes-sigs/aws-load-balancer-controller/tree/main/docs/install[review the policies on GitHub] and download the appropriate policy for your region.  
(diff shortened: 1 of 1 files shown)


latest/ug/networking/lbc-helm.adoc
```diff
@@ -43,13 +43,13 @@ Before proceeding with the configuration steps on this page, consider the follow
 
 [#lbc-helm-iam]
 == Step 1: Create IAM Role using `eksctl`
-The following steps refer to the {aws} Load Balancer Controller *v2.12.0* release version. For more information about all releases, see the https://github.com/kubernetes-sigs/aws-load-balancer-controller/releases/[{aws} Load Balancer Controller Release Page] on GitHub.
+The following steps refer to the {aws} Load Balancer Controller *v2.13.0* release version. For more information about all releases, see the https://github.com/kubernetes-sigs/aws-load-balancer-controller/releases/[{aws} Load Balancer Controller Release Page] on GitHub.
```
echapman13, line 46: Rather than incrementing with each new version, I suggest we update to the following:
"The following steps refer to the {aws} Load Balancer Controller *v2.12.0* release version. Replace *v2.12.0* with the {aws} Load Balancer Controller version you intend to deploy. For more information about all releases, see the https://github.com/kubernetes-sigs/aws-load-balancer-controller/releases/[{aws} Load Balancer Controller Release Page] on GitHub."

fincd-aws: True but a code snippet I can copy-paste-run is more useful than "here's the pattern, you need to modify it, then run"

I'll accept this PR as is, and then switch the links to the `main` branch so they stay working. I'll add a note similar to yours but more of "in production, pin the helm install to a version number and incrementally upgrade"

fincd-aws (approved):
Bumping to 2.13.0 looks good.

I will work on a better solution.

latest/ug/networking/lbc-helm.adoc
```diff
@@ -110,6 +110,7 @@ helm install aws-load-balancer-controller eks/aws-load-balancer-controller \
   --set clusterName=my-cluster \
   --set serviceAccount.create=false \
   --set serviceAccount.name=aws-load-balancer-controller
+  --version 1.13.0
```
guessi, line 113: @fincd-aws might need to add missing `\` at the end back later.

marcleblanc2: @Ahmad-RW @guessi @fincd-aws should this be --version 1.13.0, or 2.13.0?

Does this doc site have a macro for, ex. {current-version}, so the docs get auto updated after each new release?

fincd-aws: No, there's no way to macro versions like that - We keep the eksctl version in a macro but that's it.

I assume that should be 2.13.0 - the helm chart version should match the github release version. Will fix both the `\` and the version here.

The next thing is either to make this use `/mainline/` in GitHub URLs and implied latest version in Helm, or I should increment the versions more often.

marcleblanc2: I saw the PR to fix the \ in https://github.com/awsdocs/amazon-eks-user-guide/pull/1007

So I opened PR https://github.com/awsdocs/amazon-eks-user-guide/pull/1009 to fix the version number, but found a different issue, to be discussed in PR 1009.

marcleblanc2: Issue described [here](https://github.com/awsdocs/amazon-eks-user-guide/pull/1009#issuecomment-2877630951).

PR #1007, opened 2025-05-13
Update lbc-helm.adoc

Muskan-Agrawal07:
Added a backslash at the end of line 112

*Issue #, if available:*

*Description of changes:*
The helm installation command in pt.3 [here](https://docs.aws.amazon.com/eks/latest/userguide/lbc-helm.html#lbc-helm-install) is missing a backslash `\` after "--set serviceAccount.name=aws-load-balancer-controller" resulting in an error like "command not found: --version". Adding a backslash to avoid this issue.


geoffcline:
Thank you!

PR #1013, opened 2025-05-16
improve the description of the behavior of minimal update strategy

heisenbergye:
*Issue #, if available:*

improve the description of the behavior of minimal update strategy

*Description of changes:*

modify
```
1. It cordons all nodes of the node group in the beginning, so that the service controller doesn't send any new requests to these nodes. 
```

delete  
```
3. It cordons the node after every Pod is evicted and waits for 60 seconds. This is done so that the service controller doesn’t send any new requests to this node and removes this node from its list of active nodes.
```


pgasca (approved):
I agree that saying cordoning happens in two separate steps seems misleading, restructure of info looks good.

PR #1014, opened 2025-05-19
Update metrics-server.adoc

antmatts:
It's common customers get confused on what ports Metrics Server uses, especially when deploying them on Fargate nodes. I added a Considerations section to try and help clarify what to do when deploying Metrics Server on Fargate Nodes and what ports need to be allowed via Security Groups and Network ACLs for the Metrics Server to work.

*Issue #, if available:* N/A

*Description of changes:*
Added a **Considerations** section to clarify what to consider when deploy the k8s Metrics Server solution onto Fargate nodes and what ports need to be allowed for this solution to work.


pgasca (approved):
Approved, will do minor rewrites commit on top.

PR #1027, opened 2025-06-05
List Seperation, Adjustment of name of Addon

cdirubbio:
*Issue #, if available:*

*Description of changes:* 

Fixes
----------
Changed `name of addon` to `Amazon VPC CNI` in two places.

In `Step 4: Configure your cluster to use Kubernetes network policies` -> `Amazon EKS add-on`, separated Console steps from CLI Command. Currently it displays as:
```
1. ....
2. ...
3. ...
4. ...
5. ...
1. Run the following {aws} CLI command.
```
Changed to:
```
1. ...
2. ...
3. ...
4. ...
5. ...
...
You can alternatively run the following {aws} CLI command.
```


latest/ug/networking/cni-network-policy-configure.adoc
```diff
@@ -208,8 +208,7 @@ image::images/console-cni-config-network-policy.png[{aws-management-console} sho
 [discrete]
 === {aws} CLI
 
-. Run the following {aws} CLI command. Replace `my-cluster` with the name of your cluster and the IAM role ARN with the role that you are using.
-+
+You can alternatively run the following {aws} CLI command. Replace `my-cluster` with the name of your cluster and the IAM role ARN with the role that you are using.
```
fincd-aws: Little style thing: if we want "Additionally", "Optionally", "Alternatively", put it at the beginning of the sentence.

I do agree that we don't connect our sections to each other very well, but probably put the connective sentence above the first step, rather than inside it.

Would be good to standardize on an intro sentence on top of every procedure, maybe something like:

> Or, you can use the {aws} CLI using the following steps:

I don't try too hard to keep syllables below 4 per word, but this is an easy one to swap.

fincd-aws (approved):
Nice catch, these steps always differ just enough that templating them isn't really saving us time updating them.

LGTM, will test and merge.

latest/ug/networking/cni-network-policy-configure.adoc
cdirubbio: ACK, want me to commit changes for this ^?

fincd-aws: either way - if you wouldn't mind and have an idea, go for a new version.
Or I can.  I'll wait till morning to merge if you want a shot at it.

cdirubbio: Ill take a shot at it, thanks!

cdirubbio: Updated section to have CLI instructions first. Also made 2 more changes for `name of addon` where it needed to be replaced to Amazon VPC CNI

fincd-aws (approved):
LGTM

PR #1032, opened 2025-06-07
Update worker.adoc to include Note on IMTs

cdirubbio:
*Issue #, if available:*

*Description of changes:* Adds the following snippet to the docs on Self Managed Nodes:

Tags in Amazon EC2 Instance Metadata Service (IMDS) are not compatible with EKS nodes. When Instance Metadata Tags are enabled, the use of forward slashes ('/') in tag values is prevented. This limitation can cause instance launch failures, particularly when using node management tools like Karpenter or Cluster Autoscaler, as these services rely on tags containing forward slashes for proper functionality.

### DIFF
--- a/latest/ug/nodes/worker.adoc
+++ b/latest/ug/nodes/worker.adoc
@@ -27,6 +27,12 @@ To add self-managed nodes to your Amazon EKS cluster, see the topics that follow
 |`kubernetes.io/cluster/<cluster-name>`|`owned`
 |===
 
+[IMPORTANT]
+====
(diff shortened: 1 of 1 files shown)


latest/ug/nodes/worker.adoc
```diff
@@ -27,6 +27,12 @@ To add self-managed nodes to your Amazon EKS cluster, see the topics that follow
 |`kubernetes.io/cluster/<cluster-name>`|`owned`
 |===
 
+[IMPORTANT]
+====
+Tags in Amazon EC2 Instance Metadata Service (IMDS) are not compatible with EKS nodes. When Instance Metadata Tags are enabled, the use of forward slashes ('/') in tag values is prevented. This limitation can cause instance launch failures, particularly when using node management tools like Karpenter or Cluster Autoscaler, as these services rely on tags containing forward slashes for proper functionality.
```
chrisnegus, line 32: Are you sure that the first sentence is correct? I saw some indications that tags including forward slashes would not work, but that other tags that didn't have incompatibilities with the characters in those tags might work.

chrisnegus (commented):
I just have one question.

chrisnegus:
lgtm

PR #1036, opened 2025-06-11
Hybrid Nodes fix Cilium BGP ASN descriptions

csplinter:
*Description of changes:*
We previously had localASN set to on-prem router ASN, but this is incorrect. This PR fixes this description to the following.

- Configure `localASN` with the ASN for the nodes running Cilium. 
- Configure `peerASN` with the ASN for your on-premises router. 
- Configure the `peerAddress` with the on-premises router IP that each node running Cilium will peer with. 


fincd-aws (approved):
Makes sense. LGTM

PR #1047, opened 2025-06-21
Update pod-id-role.adoc

antmatts:
Adding additional guidance for how customers using Pod Identity can further restrict  a Pod Identity IAM Role trust policy by clarifying what Request Tags the AssumeRole calls provide and including an example.

*Issue #, if available:* No issue number

*Description of changes:* Customers have been asking how to restrict Pod Identity IAM Roles to ensure only specific EKS cluster, namespaces, and/or ServiceAccounts can assume the role. The default Trust Policy is very open in comparison to IRSA and we don't have any good public documentation to help clarify to customers how to achieve this. Validated and tested the RequestTags and syntax to confirm they will restrict the IAM Role access further then the default Trust Policy.

### DIFF
--- a/latest/ug/manage-access/aws-access/pod-id-role.adoc
+++ b/latest/ug/manage-access/aws-access/pod-id-role.adoc
@@ -38,17 +38,9 @@ EKS Pod Identity uses `AssumeRole` to assume the IAM role before passing the tem
 EKS Pod Identity uses `TagSession` to include _session tags_ in the requests to {aws} STS.
 
 *Setting Conditions*::
-You can use these tags in the _condition keys_ in the trust policy to restrict which service accounts, namespaces, and clusters can use this role. When the Pod Identity IAM Role is assumed, it sends the following Request Tags:
-
-* `eks-cluster-arn`
-* `eks-cluster-name`
-* `kubernetes-namespace`
-* `kubernetes-service-account`
-* `kubernetes-pod-name`
-* `kubernetes-pod-uid`
-
+You can use these tags in the _condition keys_ in the trust policy to restrict which service accounts, namespaces, and clusters can use this role. For the list of request tags that Pod Identity adds, see <<pod-id-abac-tags>>.
 +
-For example, to restrict a Pod Identity IAM Role to a specific `ServiceAccount` and `Namespace`, the following Trust Policy with the added `Condition` policies can further restrict what can assume the role:
+For example, you can restrict which pods can assume the role a Pod Identity IAM Role to a specific `ServiceAccount` and `Namespace` with the following Trust Policy with the added `Condition`:
 [source,json,subs="verbatim,attributes"]
 ----
 {
--- a/latest/ug/manage-access/aws-access/pod-id-role.adoc
+++ b/latest/ug/manage-access/aws-access/pod-id-role.adoc
@@ -36,7 +36,47 @@ EKS Pod Identity uses `AssumeRole` to assume the IAM role before passing the tem
 
 *`sts:TagSession`*::
 EKS Pod Identity uses `TagSession` to include _session tags_ in the requests to {aws} STS.
+
+*Setting Conditions*::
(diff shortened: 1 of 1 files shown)


latest/ug/manage-access/aws-access/pod-id-role.adoc
```diff
@@ -36,7 +36,47 @@ EKS Pod Identity uses `AssumeRole` to assume the IAM role before passing the tem
 
 *`sts:TagSession`*::
 EKS Pod Identity uses `TagSession` to include _session tags_ in the requests to {aws} STS.
+
+*Setting Conditions*::
+You can use these tags in the _condition keys_ in the trust policy to restrict which service accounts, namespaces, and clusters can use this role. When the Pod Identity IAM Role is assumed, it sends the following Request Tags:
+
+* `eks-cluster-arn`
```
fincd-aws: This is a copy of this list

https://github.com/awsdocs/amazon-eks-user-guide/blob/af3399368b0a1d8f6569d0b7d6977c4c22dfe16b/latest/ug/manage-access/aws-access/pod-id-abac.adoc?plain=1#L57

Its a little easier for us to only have one copy of a list, but if its relevant to the example, that's OK.

latest/ug/manage-access/aws-access/pod-id-role.adoc
```diff
@@ -36,7 +36,47 @@ EKS Pod Identity uses `AssumeRole` to assume the IAM role before passing the tem
...
+* `kubernetes-service-account`
+* `kubernetes-pod-name`
+* `kubernetes-pod-uid`
+
 +
-You can use these tags in the _condition keys_ in the trust policy to restrict which service accounts, namespaces, and clusters can use this role.
-+
-For a list of Amazon EKS condition keys, see link:service-authorization/latest/reference/list_amazonelastickubernetesservice.html#amazonelastickubernetesservice-policy-keys[Conditions defined by Amazon Elastic Kubernetes Service,type="documentation"] in the _Service Authorization Reference_. To learn which actions and resources you can use a condition key with, see link:service-authorization/latest/reference/list_amazonelastickubernetesservice.html#amazonelastickubernetesservice-actions-as-permissions[Actions defined by Amazon Elastic Kubernetes Service,type="documentation"].
\ No newline at end of file
+For example, to restrict a Pod Identity IAM Role to a specific `ServiceAccount` and `Namespace`, the following Trust Policy with the added `Condition` policies can further restrict what can assume the role:
```
fincd-aws: The double preposition is a tad awkward: For example, To restrict ...,

Writing around the second one, maybe more like:

> For example, you can restrict which pods can assume the role a Pod Identity IAM Role to a specific `ServiceAccount` and `Namespace` with the following Trust Policy with the added `Condition`:

fincd-aws (dismissed):
Thanks for this!

Just a couple nits if you don't mind.

latest/ug/manage-access/aws-access/pod-id-role.adoc
geoffcline: @antmatts do you think having it in the linked location is sufficient? or worth duplicating here?

antmatts: @geoffcline I think it's okay having it use the linked location so the list is maintained in one spot; but leave the example so users know how to use them in the Trust Policy.

antmatts: Side question, I don't see the referenced [Pod ID ABAC](https://github.com/awsdocs/amazon-eks-user-guide/blob/af3399368b0a1d8f6569d0b7d6977c4c22dfe16b/latest/ug/manage-access/aws-access/pod-id-abac.adoc) doc showing in the public documentation. Is this doc supposed to be live already?

chrisnegus: @antmatts, in the comment above, @fincd-aws referenced the file to edit. It is published here: https://docs.aws.amazon.com/eks/latest/userguide/pod-id-abac.html

PR #1058, opened 2025-07-10
Update service-accounts.adoc

liwadman:
guidance on using pod identities, STS throttling

*Issue #, if available:*
N/A Slack me.

*Description of changes:*
IRSA has a very sharp edge that lets customers get themselves throttled by STS a lot. It's time we get more opinionated about using Pod identity over IRSA, and tell customers about this in the comparison table.


geoffcline:
Thanks @liwadman ! I made some small edits. I'm sending to the PM to review the "recommends" block

PR #1061, opened 2025-07-14
Fix false statements in `eks-add-ons.adoc` & `community-addons.adoc`

0xabdi:
*Issue #, if available:*

*Description of changes:*

This PR fixes a typo


pgasca (approved):
Thank you for pointing out this mistake which is more than just a typo but the opposite of the truth.

PR #1084, opened 2025-08-06
Update private-clusters.adoc

naqvi3:
updated ssm endpoint requirement for private cluster for  ssm:RegisterManagedInstance api event as this is essential for nodes to join the cluster

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/latest/ug/clusters/private-clusters.adoc
+++ b/latest/ug/clusters/private-clusters.adoc
@@ -94,6 +94,8 @@ We recommend that you link:vpc/latest/privatelink/interface-endpoints.html#enabl
 |Amazon EKS Auth (required when using Pod Identity associations)
 |com.amazonaws.[.replaceable]`region-code`.eks-auth 
 
+|Amazon EKS SSM (required for ssm:RegisterManagedInstance API event)
+|com.amazonaws.[.replaceable]`region-code`.ssm 
 
 |Amazon EKS
 |com.amazonaws.[.replaceable]`region-code`.eks


latest/ug/clusters/private-clusters.adoc
```diff
@@ -94,6 +94,8 @@ We recommend that you link:vpc/latest/privatelink/interface-endpoints.html#enabl
 |Amazon EKS Auth (required when using Pod Identity associations)
 |com.amazonaws.[.replaceable]`region-code`.eks-auth 
 
+|Amazon EKS SSM (required for ssm:RegisterManagedInstance API event)
```
fincd-aws, line 97: The SSM agent is installed and starts with every boot, but it isn't part of EKS. EKS doesn't use it for anything.

you can register the EKS nodes with SSM and use SSM Session Manager or Run Command to do OS-level maintenance and modifications, instead of SSH. SSH is optional too, so you can have neither and still have working nodes.

fincd-aws: I'd say something like:

`|(Optional) {aws} SSM (to use SSM for node management tasks. Alternative to SSH)

fincd-aws (changes requested):
SSM is optional. Let's describe what its useful for, not the log message you clear by including it.

latest/ug/clusters/private-clusters.adoc
naqvi3: Just for reference error, ssm agent is required to run api calls for RegisterManagedInstance

Aug  5 08:02:42 ip-X.x.x.x amazon-ssm-agent[4679]: 2025-08-05 08:02:42.4312 ERROR [Registrar] failed to register identity: error calling RegisterManagedInstance API: RequestError: send request failed
Aug  5 08:02:42 ip-x.x.x.x amazon-ssm-agent[4679]: caused by: Post "https://ssm.us-east-1.amazonaws.com/": dial tcp x.x.x.x:443: i/o timeout

api doc reference: https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-setting-up-messageAPIs.html

naqvi3: @fincd-aws  I was thinking to add `|(Optional) {aws} SSM (to use SSM for node management tasks. Alternative to SSH)

however other private api endpoints such as xray are present on the doc, which are also optional but still not specified that way, so I followed the same format

fincd-aws: Ok gotcha, I'll accept as-is and add (Optional) to all of the optional ones myself.

fincd-aws (approved):
I'll follow up with additional text.

PR #1091, opened 2025-08-14
Fix wording in hybrid nodes networking doc

ahreehong:
*Issue #, if available:*

*Description of changes:*
Updating hybrid nodes networking docs to use inbound/outbound rather than ingress/egress when describing security group rules. Also added a note clarifying that remote network specific outbound rules are not auto created. (By default, EKS creates security group with a rule to enable all outbound traffic)

### DIFF
--- a/latest/ug/nodes/hybrid-nodes-networking.adoc
+++ b/latest/ug/nodes/hybrid-nodes-networking.adoc
@@ -371,7 +371,7 @@ aws ec2 associate-route-table --route-table-id [.replaceable]`RT_ID` --subnet-id
 [#hybrid-nodes-networking-cluster-sg]
 == Cluster security group configuration
 
-The following access for your EKS cluster security group is required for ongoing cluster operations. Amazon EKS automatically creates the required *ingress* security group rules for hybrid nodes when you create or update your cluster with remote node and pod networks configured. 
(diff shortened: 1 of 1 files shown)


latest/ug/nodes/hybrid-nodes-networking.adoc
```diff
@@ -371,7 +371,7 @@ aws ec2 associate-route-table --route-table-id [.replaceable]`RT_ID` --subnet-id
 [#hybrid-nodes-networking-cluster-sg]
 == Cluster security group configuration
 
-The following access for your EKS cluster security group is required for ongoing cluster operations. Amazon EKS automatically creates the required *ingress* security group rules for hybrid nodes when you create or update your cluster with remote node and pod networks configured. 
+The following access for your EKS cluster security group is required for ongoing cluster operations. Amazon EKS automatically creates the required *inbound* security group rules for hybrid nodes when you create or update your cluster with remote node and pod networks configured. Note that specific outbound rules are not auto-generated, but by default the EKS cluster security group allows all outbound traffic. If limiting outbound traffic, users are required to ensure the necessary outbound rules are configured.
```
fincd-aws: I think we should make the opposites in sentence 2 & 3 more obvious, maybe replace the "Note that..." sentence with

> Because security groups allow all outbound traffic by default, Amazon EKS doesn’t automatically modify the outbound rules of the cluster security group for hybrid nodes.

Replacing ingress with inbound matches AWS Style, and avoids confusion with K8s Ingresses, so that part is great.

fincd-aws (approved):
LGTM

also reviewed offline. I think this way, we make the different behavior for inbound . outbound more obvious.

PR #1104, opened 2025-08-20
Add package manager endpoints access to install requirements

achen1210:
*Issue #, if available:*

*Description of changes:*
Adds a note that package manager endpoints need to be allowlisted during hybrid nodes installation and upgrade process.

### DIFF
--- a/latest/ug/nodes/hybrid-nodes-networking.adoc
+++ b/latest/ug/nodes/hybrid-nodes-networking.adoc
@@ -63,6 +63,8 @@ If you are _not_ able to make your pod networks routable on your on-premises net
 
 You must have access to the following domains during the installation process where you install the hybrid nodes dependencies on your hosts. This process can be done once when you are building your operating system images or it can be done on each host at runtime. This includes initial installation and when you upgrade the Kubernetes version of your hybrid nodes.
 
+Some packages are installed using the OS's default package manager. For AL2023 and RHEL, the `yum` command is used to install `containerd`, `ca-certificates`, `iptables` and `amazon-ssm-agent`. For Ubuntu, `apt` is used to install `containerd`, `ca-certificates`, and `iptables`, and `snap` is used to install `amazon-ssm-agent`.
+
 [%header,cols="4"]
 |===
 |Component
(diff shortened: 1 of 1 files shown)


latest/ug/nodes/hybrid-nodes-networking.adoc
```diff
@@ -109,13 +109,20 @@ You must have access to the following domains during the installation process wh
...
 |===
 
 [NOTE]
 ====
 ^1^ Access to the {aws} SSM endpoints are only required if you are using {aws} SSM hybrid activations for your on-premises IAM credential provider.
 
-^2^ Access to the {aws} IAM endpoints are only required if you are using {aws} IAM Roles Anywhere for your on-premises IAM credential provider. 
+^2^ Access to the {aws} IAM endpoints are only required if you are using {aws} IAM Roles Anywhere for your on-premises IAM credential provider.
+
+^3^ For AL2023 and RHEL, `yum/dnf` is used to install `containerd`, `ca-certificates`, `iptables` and `amazon-ssm-agent`. For Ubuntu, `apt` is used to install `containerd`, `ca-certificates`, and `iptables`, and `snap` is used to install `amazon-ssm-agent`.
```
fincd-aws: > `dnf` (or `yum` respectively) is used

Forward slashes in text break some automation that we use.
```suggestion
^3^ For AL2023 and RHEL, `dnf` (or `yum` respectively) is used to install `containerd`, `ca-certificates`, `iptables` and `amazon-ssm-agent`. For Ubuntu, `apt` is used to install `containerd`, `ca-certificates`, and `iptables`, and `snap` is used to install `amazon-ssm-agent`.
```

latest/ug/nodes/hybrid-nodes-networking.adoc
```diff
@@ -109,13 +109,20 @@ You must have access to the following domains during the installation process wh
 |\https://rolesanywhere.[.replaceable]`region`.amazonaws.com
 |HTTPS
 |443
+
+|Package Manager Endpoints ^3^
```
fincd-aws: >|Operating System package manager endpoints ^3^

Be specific that we aren't adding other package managers.

```suggestion
|Operating System package manager endpoints ^3^
```

latest/ug/nodes/hybrid-nodes-networking.adoc
```diff
@@ -109,13 +109,20 @@ You must have access to the following domains during the installation process wh
...
+|Package repository endpoints are OS-specific and may vary by geographic region.
```
fincd-aws: Replace "may" with "might", for style. "May" ends up translated as the month sometimes, "might" has fewer definitions (still several, but its our style).
```suggestion
|Package repository endpoints are OS-specific and might vary by geographic region.
```

latest/ug/nodes/hybrid-nodes-networking.adoc
fincd-aws: Let's copy this info up into the paragraph at Line 64 too, so the list of packages for each OS isn't only in a footnote.

fincd-aws (dismissed):
One new piece, a couple nits you can accept.

latest/ug/nodes/hybrid-nodes-networking.adoc
achen1210: Sounds good, made the changes from above as well and moved this section. For yum/dnf I reworded to "the `yum` command is used to install...". It's a little tricky since we use the `yum` command in our code, but it seems starting with RHEL 8 and with AL2023 its just [essentially an alias/symlink for the `dnf` package manager](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/considerations_in_adopting_rhel_8/software-management_considerations-in-adopting-rhel-8#package-management-with-yum-dnf_notable-changes-to-the-yum-stack). I think we probably could change this in our code but not sure it's better since the `yum` command is well-known and remains supported.

PR #1113, opened 2025-09-12
Doc changes: add helm flags when installing aws load balancer controllerm on managed node group to avoid IMDS denied error

rookieeew:
*Issue # when deploy aws load balancer controller on Managed node groups, the controller will encounter IMDS as well if the node's hop limit is 1*

*Description of changes: delete "or if you’re deploying to Fargate or Amazon EKS Hybrid Nodes," *


chrisnegus (approved):
lgtm

PR #1115, opened 2025-09-17
Update EBS CSI Doc on self-managed policy

philljie:
Attaching an AWS managed policy to the role when installing the EBS CSI Add-on (or any add-on) is not mandatory. Users can define and attach their own custom policies to the add-on role. This PR updates the wording to make that explicit.


geoffcline:
@philljie did a little revision, lmk if you have any concerns

PR #1116, opened 2025-09-17
DOC: Added the description of the changes in access policies regarding WebSocket-related subresources

anibal2222:
*Issue #, if available:* NA

*Description of changes:* 

Starting on Kubernetes 1.31, Kubernetes uses the WebSocket protocol instead of SPDY for streaming. Following this transition, the subresources "pods/attach", "pods/exec", "pods/portforward", "pods/proxy" and "nodes/proxy" were accessible through a "GET" with Upgrade header instead of only "CREATE". As a result, this allowed users with read-only access to be able to execute streaming commands, like "exec" and "port-forward", exposing a significant security gap.

In order to prevent this behavior, the only access policies with permissions for the subresources ("pods/attach", "pods/exec", "pods/portforward", "pods/proxy" and "nodes/proxy") are: "AmazonEKSAdminPolicy", "AmazonEKSClusterAdminPolicy" and "AmazonEKSEditPolicy".

Currently, this change is not documented, which can affect the customer experience since the documentation does not reflect the current behavior of EKS clusters setting wrong expectations.


fincd-aws:
Hi thanks!
I'm also going to recheck the tables of access policies in https://docs.aws.amazon.com/eks/latest/userguide/access-policy-permissions.html

PR #1119, opened 2025-09-25
Update cross-account-access.adoc

himaatluri:
Removed some weird (at least for me) text in the policy doc examples.

Weird text removed - "&TCX5-2025-waiver;"

*Issue #, if available:*

*Description of changes:*
Removed some weird (at least for me) text in the policy doc examples.

Weird text removed - "&TCX5-2025-waiver;"

### DIFF
--- a/latest/ug/manage-access/aws-access/cross-account-access.adoc
+++ b/latest/ug/manage-access/aws-access/cross-account-access.adoc
@@ -44,7 +44,7 @@ In this example, Account B creates an IAM policy with the permissions to give to
 [source,json,subs="verbatim,attributes"]
 ----
 {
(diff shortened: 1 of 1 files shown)


himaatluri:
I see similar things on the other docs too, not sure if this is something that is normal but I'm not aware. If it is not, I'm happy to fix other places too (e.g: https://docs.aws.amazon.com/eks/latest/userguide/associate-service-account-role.html) . Thank you!

pgasca:
Thank you, the &TCX5-2025-waiver; was accidentally added, thank you for the notice. The {tcx5-waiver} is intentionally added for now as a workaround for a build issue (the rendered HTML doesn't show it though). We've addressed this throughout guide.

PR #1121, opened 2025-09-29
fix: include error output in the related NOTE admonition

nekomachi-touge:
*Issue #, if available:*
(none)

*Description of changes:*
This fix makes the error example for inaccessible EKS OIDC service endpoint on enable-iam-roles-for-service-accounts.adoc included in the related `NOTE` admonition.


pgasca:
Good catch, thank you!

PR #1136, opened 2025-10-14
[AL2 EOS Prep] Update troubleshooting.adoc, remove `Too Many Requests` section

cdirubbio:
*Issue #, if available:*

*Description of changes:* AL2023 uses nodeadm which does not make describeCluster api call. Thus, there should no longer be an issues with being throttled by nodes making this request.


cdirubbio:
Closed after discussion over slack

PR #1142, opened 2025-10-16
Remove note about IAM role usage for Windows nodes

devanshpoplii:
*Issue #, if available:*

*Description of changes:* The statement says that you can't use the same IAM role for Linux and Windows - this is incorrect. 
You you can use the same IAM role for Linux and Windows - scenario: Instead of relying on access entries, I can keep one entry in aws-auth configmap and that should work for me. 

Instead, we can say that it is recommended to create a separate IAM role. 


chrisnegus (approved):
lgtm

PR #1152, opened 2025-10-31
Added command and rectified document

HarinAcharya:
*Issue #, if available:*

- Current kubecost document has command to expose kubecost deployment when installed via helm. 
- If a user has installed the kubecost deployment via Amazon EKS add-on, then the command given in the current document fails as when kubecost is installed via addon `cost-analyzer` deployment is created.

*Description of changes:*

- Hence have added the command to expose kubecost deployment when installed via addon and rectified the wordings accordingly


tucktuck9 (approved):
LGTM

Issue #1202, opened 2026-01-07
Update kubernetes-versions RSS URL


ChrisEke:
The URL for configuring RSS feed for kubernetes-versions in [https://github.com/awsdocs/amazon-eks-user-guide/blob/mainline/latest/ug/versioning/kubernetes-versions.adoc](https://github.com/awsdocs/amazon-eks-user-guide/blob/mainline/latest/ug/versioning/kubernetes-versions.adoc?plain=1#L64C1-L64C115) should be updated:

```diff
- https://github.com/awsdocs/amazon-eks-user-guide/commits/mainline/latest/ug/clusters/kubernetes-versions.adoc.atom
+ https://github.com/awsdocs/amazon-eks-user-guide/commits/mainline/latest/ug/versioning/kubernetes-versions.adoc.atom
```

jlbutler:
Merged a fix for this, thanks for the issue @ChrisEke and thanks for the PR @MatheusFlausino

PR #1204, opened 2026-01-07
Fix broken RSS feed URLs after path change

MatheusFlausino:
*Issue #, if available:* No

*Description of changes:*

When kubernetes-versions.adoc and platform-versions.adoc were moved from latest/ug/clusters/ to latest/ug/versioning/, the RSS feed URLs in both files were not updated. This broke the feeds because they still pointed to the old paths.

*Solution*
Updated the RSS feed URLs in both files to match the new paths:
kubernetes-versions.adoc: Changed from /clusters/kubernetes-versions.adoc.atom to /versioning/kubernetes-versions.adoc.atom
platform-versions.adoc: Changed from /clusters/platform-versions.adoc.atom to /versioning/platform-versions.adoc.atom

### DIFF
--- a/latest/ug/versioning/kubernetes-versions.adoc
+++ b/latest/ug/versioning/kubernetes-versions.adoc
@@ -61,7 +61,7 @@ To receive notifications of all source file changes to this specific documentati
 
 [source,none]
 ----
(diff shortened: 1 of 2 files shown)


jlbutler:
fixes #1202

jlbutler:
thanks @MatheusFlausino!

PR #1212, opened 2026-01-22
Specifies DeletePropagationPolicy and clarifies region

doddstr13:
DeletePropagationPolicy is required for the capability definition and the instructions should accommodate for IAM Identity Center in a different region from the EKS cluster.

*Issue #, if available: NA

*Description of changes: Updates ArgoCD Capabilities installation page for eksctl to specify delete propagation policy and clarify instructions for when IDC is in a different region than the cluster.


srhsrhsrhsrh:
Hi @doddstr13, can you please rebase on current mainline? It seems like conflict resolution chose the whole file instead of just the single change. Thanks!

PR #1222, opened 2026-03-12
Correcting the storageready repair action

sumanthst25:
*Issue #, if available:*

*Description of changes:*

According to the NMA documentation : https://docs.aws.amazon.com/eks/latest/userguide/node-health-nma.html#node-health-Storage and NMA source code : https://github.com/aws/eks-node-monitoring-agent/blob/main/pkg/reasons/reasons.yaml#L341-L367, there is no "condition" for StorageReady that can trigger the replace/reboot of the instance as all the scenarios are events.

Therefore I am creating this PR to correct the docs which say that StorageReady triggers a replace in 30 mins.

### DIFF
--- a/latest/ug/nodes/node-health-nma.adoc
+++ b/latest/ug/nodes/node-health-nma.adoc
@@ -478,7 +478,7 @@ The monitoring condition is `NetworkingReady` for issues in the following table 
 [#node-health-Storage]
 == Storage node health issues
 
-The monitoring condition is `StorageReady` for issues in the following table that have a severity of “Condition”.
+The monitoring condition is `StorageReady` for issues in the following table that have a severity of “Event”.
 
 [%header,cols="4"]
 |===
--- a/latest/ug/nodes/node-repair.adoc
+++ b/latest/ug/nodes/node-repair.adoc
@@ -56,8 +56,8 @@ For a detailed list of node health issues detected by the EKS node monitoring ag
 
 |StorageReady
 |StorageReady indicates whether the node's storage subsystem is functioning correctly (disks, filesystems, I/O).
-|30m
-|Replace
+|N/A
+|None
 
 |Ready
 |Ready is the standard Kubernetes condition indicating the node is healthy and ready to accept pods.


sumanthst25 (commented):
Correcting the documentation to reflect StorageReady of NMA conditions as Events

sumanthst25:
@prasad0896 Adding you to review this.

latest/ug/nodes/node-health-nma.adoc
```diff
@@ -478,7 +478,7 @@ The monitoring condition is `NetworkingReady` for issues in the following table
 [#node-health-Storage]
 == Storage node health issues
 
-The monitoring condition is `StorageReady` for issues in the following table that have a severity of “Condition”.
+The monitoring condition is `StorageReady` for issues in the following table that have a severity of “Event”.
```
prasad0896, line 481: This should be `Condition` only. NMA will only change the status on `condition`. Subsequently only conditions are considered for any repair action.

latest/ug/nodes/node-repair.adoc
```diff
@@ -56,8 +56,8 @@ For a detailed list of node health issues detected by the EKS node monitoring ag
 
 |StorageReady
 |StorageReady indicates whether the node's storage subsystem is functioning correctly (disks, filesystems, I/O).
-|30m
-|Replace
+|N/A
```
prasad0896, line 59: Thanks for the change, however I do want to callout that if we ever add a reason to NMA that `has DefaultSeverity: Fatal,` then repair will be triggered on the node in default cases.

latest/ug/nodes/node-health-nma.adoc
BrianS99: @sumanthst25 could you please address these comments

sumanthst24: None of the current StorageReady scenarios fall under "Condition". All the scenarios are listed under Events category. But this can be ignored considering the future changes that may come in

